PR #3, opened 2015-05-28
libmosquitto-based C client suggestion

jpmens:
I suggest replacing C client example with something like this from libmosquitto.


dergraf:
Thank you very much!

PR #4, opened 2015-05-28
nitpicking :-)

jpmens:


dergraf:
Could you also change the https://github.com/erlio/vernemq/blob/master/docs/index.rst with your suggestions?

jpmens:
Done.

dergraf:
Thanks!

PR #16, opened 2015-05-31
rebar3 support working with cwd rebar3 escript

tsloughter:
I had to make some changes to relx that I haven't got merged into it and rebar3 yet, so only works wtih the bundled rebar3 escript right now.

I don't like some of the changes and will have to investigate fixes.

First, it seems `relx` isn't replacing variables in the template strings within the variable, which is why I changed `runner_base_dir` to `{runner_base_dir,       "$(cd `dirname $0` ; /bin/pwd)/.."}`. Second, I had to change it to `;` because `&&` was being replaced with `&amp;&amp;`.


tsloughter:
Nevermind, the `relx` patch isn't needed. But `rebar3` script here is still "special", it has a patch for building escripts like cuttlefish I need to get merged to rebar3 repo.

dergraf:
I guess the rebar3 patch will eventually go into rebar3 master?

dergraf:
on a fresh clone of vernemq (rebar3-support branch) a ``./rebar3 release`` builds the release! :) However the cuttlefish config file ``vernemq.conf`` is not generated.. do I have to invoke the rebar3_cuttlefish plugin somehow in advance?

tsloughter:
Yea, the rebar3 patch was merged last night. So now would work with any rebar3 we release.

The cuttlefish plugin should be invoked due to `{provider_hooks, [{post, [{release, cuttlefish}]}]}.` I'll take a look in a minute.

PR #81, opened 2016-01-14
Make it build and run on erlang 18

marianoguerra:
As mentioned in https://github.com/erlio/vernemq/issues/80 

I see that the plumtree now usage is already fixed upstream https://github.com/helium/plumtree/blob/master/src/plumtree_peer_service_manager.erl#L144

on vmq_server the offending lines (on master) are:

* https://github.com/erlio/vmq_server/blob/master/src/vmq_mqtt_fsm.erl#L85
* https://github.com/erlio/vmq_server/blob/master/src/vmq_mqtt_fsm.erl#L1032

which can be fixed as in plumtree using time_compat: https://github.com/helium/plumtree/blob/master/rebar.config#L3

when those fixed are pushed to the dependencies then the overrides can be removed.
the edown dep should be removed when all the dependencies use a version that builds on erlang 18 or at least the first dependency of vernemq that depends on edown does (given dependency resolution rules from rebar3)


tsloughter:
edown should not be in the main deps list. It should be moved to the docs profile:

```erlang
{profiles, [{docs, [{deps, [edown]}]}]}.
```

Also, there is no `require_otp_vsn` in rebar3. it should be changed to `minimum_otp_vsn`.

PR #204, opened 2016-09-30
Give app_version a value in custom builds

larshesel:
This value is getting set in other ways when creating packages.

This means `vernemq version` will now return a string like
'0.14.2-03f5c05' instead of the empty string.

### DIFF
--- a/vars.config
+++ b/vars.config
@@ -29,6 +29,7 @@
 {runner_user,           ""}.
 {runner_wait_process,   "vmq_cluster_node_sup"}.
 {runner_ulimit_warn,    65536}.
+{app_version,           "$(git describe --tag --dirty)"}.
 
 %% vmq_server
 {max_connections, 10000}.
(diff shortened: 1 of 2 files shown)


dergraf:
I tried a `make rel` on your branch and got:

    _build/default/rel/vernemq/bin/vernemq start
    fatal: No annotated tags can describe '4f8b0709b5705c3cb1297b8cbff3b24ff6c4b28a'.
    However, there were unannotated tags: try --tags.

VerneMQ was successfully started, but every `vernemq` or `vmq-admin` command is printing the error message above.

larshesel:
aha, interesting - why didn't I see that? Investigating...

larshesel:
@dergraf I changed it so it will also work with lightweight tags - it will fail if no tags are present (you can clone a repo by picking for instance only the last commit, but it is unusual, so I think it's ok).

Can you try again?

larshesel:
this really isn't that important - closing.

PR #224, opened 2016-11-08
Build packages without a _checkouts folder

larshesel:
This is useful when upgrading dependencies but not wanting to commit the
changes yet.


larshesel:
Merged this as it fixes a few annoying bugs and doesn't influence other Makefile targets.

PR #228, opened 2016-11-21
Fix racing subscriber test

larshesel:
This builds on top of #227 and that PR should be merged before this.

The racing subscriber test was sometimes failing. I noticed that it was because the `vmq_subscriber:subtract/3` was sometimes called with the second argument empty (`[]`) which we don't handle. See the details in the commit.

The fix relies on the two lists to `get_changes/2` to be sorted. I haven't been able to verify if they indeed are sure, but I couldn't imagine how the implementation of `subtract` would ever work in the first place if they aren't.


larshesel:
@dergraf Oh, this doesn't cover all cases either - will have another go at it later

larshesel:
Take 2. Now you can have a look @dergraf

PR #231, opened 2016-11-24
Fix failing tests

larshesel:
This is still a WIP, but want to see if travis likes it or not


larshesel:
Alright - there are still other tests that are failing from time to time, but this fixes at least three of them. @dergraf take a look when you have time for it.

PR #232, opened 2016-11-29
Make compatible with openssl 1.1

larshesel:
I can test that this works on openssl 1.1 - @dergraf or @ioolkos can you check if this still builds and works on pre-1.1 openssl?


larshesel:
For reference the openssl 1.1 api changes: https://wiki.openssl.org/index.php/1.1_API_Changes

PR #235, opened 2016-12-05
Make it possible to rewrite client id in auth_on_register

larshesel:
Looking at erlio/vmq_webhooks#6 I noticed that we in the handling of the `auth_on_register` hook never updates the `client_id` (see https://github.com/erlio/vernemq/blob/master/apps/vmq_server/src/vmq_mqtt_fsm.erl#L600-L601)

So I created this pull request, but I'm pretty sure there are some other things that needs to be handled as well: Currently we allow overwriting both `mountpount` and `subscriber_id`, and the latter also contains the mountpoint. That is confusing, so I'd expect that we should only expose either `subscriber_id` or `mountpoint` and `client_id`.

Let's figure out what the right thing to do here is.

/cc @dergraf

### DIFF
--- a/apps/vmq_server/src/vmq_mqtt_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt_fsm.erl
@@ -63,7 +63,6 @@
 
           %% config
           allow_anonymous=false             :: boolean(),
-          mountpoint=""                     :: mountpoint(),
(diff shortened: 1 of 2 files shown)


larshesel:
Ok @dergraf , this is ready for review...

dergraf:
lgtm!

PR #237, opened 2016-12-06
Fix retain case_clause bug

larshesel:
When introducing the new retain cache/update method two cases of the old format
was missed and left behind.

Fixes erlio/vernemq#236.


dergraf:
lgtm!

PR #240, opened 2016-12-07
use vernemq_dev repo

dergraf:
@larshesel  please review


apps/vmq_commons/include/vmq_types.hrl
```diff
@@ -8,19 +9,7 @@
...
--type mountpoint()          :: string().
--type subscriber_id()       :: {mountpoint(), client_id()}.
--type reg_view()            :: atom().
--type topic()               :: [binary()] | binary().
--type qos()                 :: 0 | 1 | 2.
--type routing_key()         :: [binary()].
--type payload()             :: binary().
--type flag()                :: ?true | ?false | boolean() | empty. % empty for test purposes
--type msg_ref()             :: binary().
+-type fflag()               :: flag() | ?true | ?false | empty. % empty for test purposes
```
larshesel: So `flag()` (and all the other types) now lives in `vernemq_dev` and I see `flag()` is now a type alias for `boolean()`. Was the idea to make `flag()` a 'clean' flag type and then make `fflag()` the internal version we need which can take on a couple of other meanings/values as well? What does the prefixed `f` mean?

dergraf: yeah you're right.
`flag()` is kind of the high level type that is used at multiple places within `vmq_server`. The goal is, that we can ensure that all `vmq_server` sees is a `boolean()` and not `0|1`. However the parser deals with flags containing `0|1` as well as the conversion from `true|false -> 1|0` the `empty` type is for testing purposes.

PR #242, opened 2016-12-08
Fix another issue in the parser logic

larshesel:
As reported by @waliferus on slack:

When fixing the last parser issue, we inadvertently changed the logic so
parse/4 would be returning `packet_exceeds_max_size` when `MaxSize=0`
and `byte_size(Data) < DataSize` which of course makes no sense.

This commits makes sure we handle all six cases:

- no max, enough data -> parse
- no max, not enough data -> more
- max - enough data - max exceeded -> error max exceeded
- max - enough data - max not exceeded -> parse
- max - not enough data - max exceeded -> error max exceeded
- max - not enough data - max not exceeded -> more

@dergraf please review


dergraf:
@larshesel LGTM!

PR #247, opened 2016-12-28
VMQL

dergraf:
`VMQL` provides an SQL-like language to retrieve information from a VerneMQ node.
The `vmq-admin session list` commands now uses `vmql` to query the sessions.


dergraf:
@larshesel I think the vmq_ql code (without the repl code) is basically ready to be reviewed. However it's still WIP

apps/vmq_server/src/vmq_info_cli.erl
```diff
@@ -0,0 +1,95 @@
...
+     "    list        lists all running sessions\n",
+     "  Use --help after a sub-command for more details.\n"
+    ].
+
+vmq_session_list_usage() ->
+    Options = [io_lib:format("  --~p\n", [Item])
+               || Item <- vmq_info:session_info_items()],
+    ["vmq-admin session list\n\n",
+     "  Prints some information on running sessions\n\n",
+     "Options\n\n" | Options
```
larshesel: The new `session list` cmd shows more information than the old one. There's way too much to make sense of and the column headers a truncated which is less than helpful.

```
$ _build/default/rel/vernemq/bin/vmq-admin session list
+------+------+--------+------+-----+----+----+-----+--------+------+-------+-------+----+-----+----+
|__moun|clean_|client_i|delive|is_of|is_o|is_p|mount|  node  |num_se|offline|online_|queu|sessi|stat|
+------+------+--------+------+-----+----+----+-----+--------+------+-------+-------+----+-----+----+
|      | true |mosqsub/|fanout|false|true|fals|     |VerneMQ@|  1   |   0   |   0   |<0.4|<0.45|onli|
+------+------+--------+------+-----+----+----+-----+--------+------+-------+-------+----+-----+----+
```

I'd therefore propose that we by default only quuery a subset of the information. Something like: 

```
$ _build/default/rel/vernemq/bin/vmq-admin session list --client_id --user --is_online --statename --peer_host --peer_port
+------------------+---------+---------+---------+---------+---------+
|    client_id     |is_online|peer_host|peer_port|statename|  user   |
+------------------+---------+---------+---------+---------+---------+
|mosqsub/1840-algol|  true   |localhost|  43118  | online  |undefined|
+------------------+---------+---------+---------+---------+---------+
```

I would then change the line `Prints some information on running sessions\n\n` to something like:

```
  Prints information about running sessions.
      Default options: --client_id --user --is_online --statename --peer_host --peer_port 
```

apps/vmq_server/src/vmq_info.erl
```diff
@@ -0,0 +1,169 @@
...
+                   name =       queue_base,
+                   depends_on = [],
+                   provides =   [node, mountpoint, client_id, queue_pid],
+                   init_fun =   fun row_init/1,
+                   include_if_all = true
+                  },
+    Queues = #vmq_ql_table{
+                   name =       queues,
+                   depends_on = [QueueBase],
+                   provides = [status,
```
larshesel: AFAICT the `status` will never be part of the map returned from `vmq_queue:info/1` and will therefore always default to `null`.

apps/vmq_server/src/vmq_info.erl
```diff
@@ -0,0 +1,169 @@
...
+                               deliver_mode,
+                               queue_size,
+                               session_pid,
+                               is_offline,
+                               is_online,
```
larshesel: Maybe get rid of the `is_offline` and only have `is_online` (other states can be retrieve through `statename`)?

apps/vmq_server/src/vmq_info_cli.erl
larshesel: Fixed these above concerns in dergraf/vernemq#1

larshesel:
IT seems there is a bug where any old session is listed even after the client is not longer there (and clean_session was true):

```
$ _build/default/rel/vernemq/bin/vmq-admin session list
+-------------------+---------+----------+---------+---------+---------+
|     client_id     |is_online|mountpoint|peer_host|peer_port|  user   |
+-------------------+---------+----------+---------+---------+---------+
|mosqsub/22453-algol|  false  |          |  null   |  null   |  null   |
|        l1         |  false  |          |  null   |  null   |  null   |
|mosqsub/27348-algol|  false  |          |  null   |  null   |  null   |
|mosqsub/26197-algol|  false  |          |  null   |  null   |  null   |
|mosqsub/21399-algol|  false  |          |  null   |  null   |  null   |
|mosqsub/23259-algol|  false  |          |  null   |  null   |  null   |
|mosqsub/24197-algol|  false  |          |  null   |  null   |  null   |
|        l2         |  true   |          |localhost|  48832  |undefined|
+-------------------+---------+----------+---------+---------+---------+
```

I might try to look into this later.

apps/vmq_server/src/vmq_info.erl
larshesel: Also addressed this in dergraf/vernemq#1

apps/vmq_ql/src/vmq_ql_query_mgr.erl
```diff
@@ -0,0 +1,226 @@
...
+                        ets:insert(ResultTab, {{RowKey, AccAccIdx}, Row}),
+                        AccAccIdx + 1
+                end, AccIdx, safe_fetch(Pid, desc, Limit))
+      end, Idx, Ready),
+    State#state{idx=NewIdx}.
+
+safe_fetch(Pid, Order, Limit) ->
+    case is_process_alive(Pid) of
+        true ->
+            vmq_ql_query:fetch(Pid, Order, Limit);
```
larshesel: I don't believe the `safe_fetch` with `is_process_alive` checks is needed as the `vmq_ql_query:fetch/3` handles the `noproc` exception.

apps/vmq_ql/src/vmq_ql_query.erl
```diff
@@ -0,0 +1,380 @@
...
+            %% Raising inside an ets:fold allows us to stop the fold
+            ok;
+        E1:R1 ->
+            ets:delete(Results),
+            lager:error("Select query terminated due to ~p ~p", [E1, R1]),
+            exit({E1, R1})
+    end,
+    case is_integer(Limit) of
+        false ->
+            ok;
```
larshesel: How can `Limit` not be an integer at this point? If it is supposed to only be an integer maybe put a guard on the `vmq_ql_query:fetch/3` function as well on `vmq_ql_query_mgr:fold/4` function and skip the check here?

larshesel: Aah, I see in `vmq_ql_parser` that the limit might also be an expression?

files/vmq-admin
```diff
@@ -21,4 +21,4 @@ cd $RUNNER_BASE_DIR
 # Identify the script name
 SCRIPT=`basename $0`
 
-$NODETOOL rpc vmq_server_cli command $SCRIPT $@
+$NODETOOL rpc_infinity vmq_server_cli command $SCRIPT $@
```
dergraf: we should revert it to rpc. As the rpc_infinity I planned to use for the repl part.

larshesel:
Only issue remaining is the issue of multiple filters. For example:
```
vmq-admin session list  --is_online=true --node="VerneMQ@127.0.0.1"
```

doesn't return any results, but the same command with only one of the filters returns the same list:
```
vmq-admin session list --client_id --node --is_online
+-------------------+---------+-----------------+
|     client_id     |is_online|      node       |
+-------------------+---------+-----------------+
|mosqsub/30577-algol|  false  |VerneMQ@127.0.0.1|
|mosqsub/1242-algol |  false  |VerneMQ@127.0.0.1|
|mosqsub/3908-algol |  false  |VerneMQ@127.0.0.1|
|mosqsub/2739-algol |  false  |VerneMQ@127.0.0.1|
|mosqsub/1270-algol |  false  |VerneMQ@127.0.0.1|
|mosqsub/4270-algol |  false  |VerneMQ@127.0.0.1|
+-------------------+---------+-----------------+
```

The query above looks like this internally:
```
(VerneMQ@127.0.0.1)3> vmq_ql_parser:parse("SELECT * FROM sessions WHERE node=\"VerneMQ@127.0.0.1\" AND is_online=false LIMIT 100").
[{type,"SELECT"},
 {fields,[all]},
 {from,sessions},
 {where,[{op,node,equals,
             {op,<<"VerneMQ@127.0.0.1">>,'and',
                 {op,is_online,equals,false}}}]},
 {orderby,[]},
 {limit,100}]
```

The problem seems to be that we do not recurse when handling the operators. I was thinking adding something like:

```diff
+%% Recurse the left and right values if possible
+eval_op(Op, {op, LV1, LOp, LV2}, {op, RV1, ROp, RV2}) ->
+    eval_op(Op, eval_op(LOp, LV1, LV2), eval_op(ROp, RV1, RV2));
+eval_op(Op, L, {op, RV1, ROp, RV2}) ->
+    eval_op(Op, L, eval_op(ROp, RV1, RV2));
+eval_op(Op, {op, LV1, LOp, LV2}, R) ->
+    eval_op(Op, eval_op(LOp, LV1, LV2), R);
+
+eval_op('or', V1, V2) -> v(V1) or v(V2); 
+eval_op('and', V1, V2) -> v(V1) and v(V2); 
 eval_op(equals, V1, V2) -> v(V1) == v(V2);
 eval_op(not_equals, V1, V2) -> v(V1) /= v(V2)
```
anyway - something to look into. I think maybe we need to handle this at a higher level - it seems the `eval_query` funcs are built for and'in and or'ing individual queries elements together. for that to work I guess the query should have been on the form:

```
    [{op,node,equals,<<"VerneMQ@127.0.0.1">>},
     {'and',{op,is_online,equals,false}}]
```

Issue #230, opened 2016-11-24
suggestion: MQTT Client Load Balancing with Shared Subscriptions


dcy:
Hi!
http://www.hivemq.com/blog/mqtt-client-load-balancing-with-shared-subscriptions/
Shared Subscriptions are a non-standard-MQTT feature, but it is usefull for some use cases.
Maybe it is good for VerneMQ to support this feature.

larshesel:
Hi @dcy this is something we've been considering and may well introduce in the future. It doesn't have super high priority at the moment, but if someone would like to sponsor or contribute this feature that would be fantastic.

mmindenhall:
In case it matters, this isn't something that Hive is doing in isolation...this feature has been added to a recent working draft of the MQTT v5 standard: https://issues.oasis-open.org/browse/MQTT-234.

larshesel:
Hi all! Good news, we've been discussing this internally again and we'll be working on bringing shared subscriptions to VerneMQ in the near future ;)

perotom:
@larshesel i think this feature is highly important for distributed systems. Any updates when this feature will be available? I am really looking forward in using your broker in production with this feature.

larshesel:
Hi @perotom the feature is being worked on as we speak. With some luck it'll go into review this week.

larshesel:
Hi - we just merged this feature into master and it's available in the nightly builds. See http://vernemq.com/docs/configuration/shared_subscriptions.html and https://github.com/erlio/vernemq/blob/master/changelog.md#nightly-will-become-next-release

dcy:
@larshesel Thank for the effort.
In the document, ```currently only members of a shared subscription that are online are considered for for delivery```. Maybe there is an plugin hook to tell me this msg has no online member to dilivery, so i can handle it by another way.

larshesel:
Hi @dcy there are currently no hooks for that. As mentioned, this behaviour may change (and be made configurable). Do you need the messages to be stored in the offline queue?

dcy:
It is better to choose by configure.
Because when i send msg to another server by ```shared subscriptions```,  if there is no online clients(myabe the server restart or crash), it will lost the msg.
It is better to have offline msgs; 
Or a hook to tell me then i can send it to another topic(not shared subscriptions) or handle it by another way. 
Then the msg will not lost

larshesel:
[...]

Issue #221, opened 2016-11-04
Timeline to support consumer session balancing across different cluster nodes


hijklmno:
Hi VerneMQ team,

Is support for consumer session balancing across different cluster nodes on the roadmap? As in, if I subscribe on node 1 with clientId `subber1` and subscribe on node 2 with clientId `subber2` on the same topic, will the message be distributed between `subber1` and `subber2`?

Thanks again!

larshesel:
Hi @hijklmno, sorry for the slow response. we were working on a feature like this but we haven't had time to finish it yet, as we've had to prioritize other things. For the the moment it doesn't have too high priority, so it's hard to say when it will be completed.

May I ask why you need this feature? It is already possible to do session sharing on a single node (i.e., connecting multiple clients using the same client-id by setting the `allow_multiple_sessions = true` and `queue_deliver_mode = balance` in the vernemq.config - or for a specific client in a plugin). This should work fine for message flows up to about 20-25K msgs/s, depending on the hardware of course.

hijklmno:
No worries, thank you for your response @larshesel. 

The biggest reason is uptime. We run our infra on AWS, and for production ready systems, we require at least two running instances of anything, in different availability zones. Hosting on cloud providers biggest downside is that their servers can go down at any time, for any reason. While it is possible to only have one instance of verne with session sharing on a single node, and auto launch a new instance in case the instance goes down, the downtime in between does not fit our SLA.

dergraf:
duplicate #230

larshesel:
Hi - we just merged this feature into master and it's available in the nightly builds. See http://vernemq.com/docs/configuration/shared_subscriptions.html and https://github.com/erlio/vernemq/blob/master/changelog.md#nightly-will-become-next-release

PR #252, opened 2017-01-04
Subscriber groups/shared subscriptions

larshesel:
@dergraf Ready for review

 - policies implemented: `random` and `prefer_local`. We could consider adding `only_local` as well.
 - enqueue_many with options only implemented in 'online' state. Should probably be implemented in others as well?
 - needs to be verified against the MQTTv5 spec.
 - must be squashed before merging.

### DIFF
--- a/apps/vmq_server/src/vmq_cluster_com.erl
+++ b/apps/vmq_server/src/vmq_cluster_com.erl
@@ -142,9 +142,10 @@ process_bytes(Bytes, Buffer, St) ->
             process(BFrames, St),
             process_bytes(Rest, <<>>, St);
         _ ->
+            %% if we have received something else than "vmq-send" we
+            %% will buffer everything unbounded forever and ever!
             {ok, NewBuffer}
     end.
-
 
 process(<<"msg", L:32, Bin:L/binary, Rest/binary>>, St) ->
     case binary_to_term(Bin) of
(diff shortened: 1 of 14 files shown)


apps/vmq_server/src/vmq_cluster_com.erl
```diff
@@ -142,34 +142,58 @@ process_bytes(Bytes, Buffer, St) ->
...
+                              _:_ ->
+                                  CallerPid ! {Ref, {error, cant_remote_enqueue}}
+                          end
+                  end);
+        {CallerPid, Ref, {enqueue_many, SubscriberId, Msgs, Opts}} ->
+            spawn(fun() ->
+                          try
+                              case vmq_queue_sup_sup:get_queue_pid(SubscriberId) of
+                                  QueuePid when is_pid(QueuePid) ->
+                                      %% TODO, maybe replace with dedicated
```
dergraf: what is this?

apps/vmq_server/src/vmq_reg.erl
```diff
@@ -268,6 +268,61 @@ register_session(SubscriberId, QueueOpts) ->
...
+    publish_to_subscriber_groups(NewMsg, SGPolicy, SubscriberGroups).
+
+publish_fold_acc(Msg) -> {Msg, undefined}.
+
+publish_to_subscriber_groups(_,_, undefined) -> ok;
+publish_to_subscriber_groups(Msg, Policy, SubscriberGroups) when is_map(SubscriberGroups) ->
+    publish_to_subscriber_groups(Msg, Policy, maps:to_list(SubscriberGroups));
+publish_to_subscriber_groups(_,_, []) -> ok;
+publish_to_subscriber_groups(Msg, Policy, [{Group, []}|Rest]) ->
+    lager:info("can't publish to subscriber group ~p due to no subscriber available", [Group]),
```
dergraf: debug instead of info? If debug add the Msg to the statement.

apps/vmq_server/src/vmq_reg.erl
```diff
@@ -268,6 +268,61 @@ register_session(SubscriberId, QueueOpts) ->
...
+                    ok = vmq_queue:enqueue(QPid, {deliver, QoS, Msg}),
+                    publish_to_subscriber_groups(Msg, Policy, Rest)
+            end;
+        {Node, SubscriberId, QoS} = Sub ->
+            Term = {enqueue_many, SubscriberId, [{deliver, QoS, Msg}], #{states => [online]}},
+            case vmq_cluster:remote_enqueue(Node, Term) of
+                ok ->
+                    publish_to_subscriber_groups(Msg, Policy, Rest);
+                {error, Reason} ->
+                    lager:warning("can't publish to subscriber group on remote node ~p due to '~p'",
```
dergraf: Maybe debug instead of warning

apps/vmq_server/src/vmq_cluster_com.erl
larshesel: removed.

apps/vmq_server/src/vmq_reg.erl
larshesel: fixed and reworded.

apps/vmq_server/src/vmq_reg.erl
larshesel: fixed and reworded.

dergraf:
LGTM! :)

larshesel:
Squashed and rebased the subscriber group onto master and changed the message compatibility layer so the cluster com is compatible with pre-subscriber group nodes (need to test this)

larshesel:
Link this PR to the related issues: #230 and #221

dergraf:
I think as the official spec isn't final yet, this is good to merge, we can change it later, and mark this feature as beta. 

comments?

larshesel:
I think that's fair. I'll do some mixed-cluster testing and merge if that looks OK. Then I'll start documenting it (and include a note about the beta status etc).

larshesel:
Hi @dergraf it looks good to me - my mixed-cluster tests didnt't reveal any issues with the cluster comm. I added an intermediate entry to the changelog so we can document features as we add them and then we'd only need to change the heading when we do the actual release.

Could you take a quick look at the changelog? If the content there is OK with you I'll merge it (or feel free to do it)

ig0r:
guys,

Is it included in the release? 

Trying it in 0.15.3 and don't see it's working

ig0r:
Oh, I see. Thanks for the heads-up!

PR #256, opened 2017-01-10
Make this version compatible with subscriber groups

larshesel:
Since the subscriber groups mandates a change in the #vmq_msg{} record,
we need to be able to handle that record before introducing the actual
change to make a cluster node-by-node upgrade possible.


larshesel:
LGT @dergraf ;)

PR #261, opened 2017-01-20
Subscriber groups online first

larshesel:
This pull request makes sure that messages published to a shared subscription are first attempted to be delivered subscribers with queues that are online first and only afterwards to to non-online queues.

Further, it implements the `local_only` policy.


dergraf:
LGTM!

PR #265, opened 2017-02-02
Configure plugins via vernemq conf

larshesel:
To eliminate the vmq_plugin.conf file and make vernemq more container friendly, I thought about something in this direction. What do you think?

Changes:
 - removes the dependency on lib/vmq_plugin.conf
 - makes it possible to configure plugins via the vernemq.conf file. Example:
```
plugin.3vmq_diversity.plugin = vmq_diversity
plugin.3vmq_diversity.path = /home/lhc/dev/erl.io/vmq_diversity/_build/default
```


larshesel:
Hi @codeadict an option like that would be interesting. We currently don't have any method to update the `vernemq.conf` in place - only generate it once when building the vernemq release. So I don't have any good ideas how to do that at the moment, I'm afraid. Suggestions are welcome, of course!

larshesel:
Now depends on erlio/vmq_diversity#25 to be merged.

larshesel:
The part in `vmq_webhooks` is ready (erlio/vmq_webhooks#16) and should be reviewed and merged before this PR.

apps/vmq_server/priv/vmq_server.schema
```diff
@@ -1470,3 +1470,66 @@
...
+%% shipped with VerneMQ.
+{mapping, "plugins.$name.path", "vmq_server.user_plugins", [
+                                                            {datatype, directory},
+                                                            {commented, "/path/to/myplugin"},
+                                                            {include_default, "mypluginname"}
+                                                          ]}.
+
+%% @doc File based authentication plugin. Enabled by default. Comment
+%% out to disable.  
+{mapping, "plugins.plugin1.plugin", "vmq_server.user_plugins", [
```
dergraf: should we name it `plugins.p1.plugin` quite a lot of `plugin` in here.

dergraf: or should we even use `plugins.p1.app = vmq_acl` ??

larshesel: I started out calling it `plugins.p1.name`, but I changed it to `plugin` because I found that easier to document. I think I'd prefer `name` over `app`. Also, I could change the config name of the plugin to something else, so it would read something like: `plugins.1acl.name = vmq_acl` ? Would that be nicer to read?

dergraf: I think I'd like `name` more ;)

larshesel:
Replaced by #270

PR #270, opened 2017-02-10
The big integration

larshesel:
Replaces #265,
Relaces erlio/vmq_diversity#25
Replaces erlio/vmq_webhooks#16

- introduces the plugin enable/disable directly in the vernemq.conf
- configure vmq_acl, vmq_passwd, vmq_bridge schema changes to conform to the vernemq.conf changes
- include vmq_webhooks and vmq_diversity in the ./apps directory.
...

To check
- test that all plugins are still working as intended after the changes to the schema files.
- that the generated vernemq.conf file is (at least as) understandable (as before)

Todo: update documentation on the website.

Other?

Issue #269, opened 2017-02-07
Defining more than one bridge fails parsing the config


gokr:
### Environment

- VerneMQ Version: 0.15.3-f73cc2a0-1
- OS: Ubuntu 14.04.5 LTS
- Erlang/OTP version (if building from source):

### Expected behavior

No failure in parsing config. :)

### Actual behaviour
Define first a br0, then a br1 - it will fail parsing the config when second bridge added.

larshesel:
Hi @gokr I just confirmed this. We'll start figuring out the reason and fixing it shortly.

larshesel:
I narrowed it down to our usage of `lists:usort/1` for some bridge options and not for others which causes a badmatch in the `TCPIPs` variable. See https://github.com/erlio/vernemq/blob/master/apps/vmq_bridge/priv/vmq_bridge.schema#L346.

@dergraf do you remember why `lists:usort/1` was applied there? For the uniqueness property? If so I don't see why we'd need that, but I might be missing something. I'd propose removing the `lists:usort/1` calls (which makes it work).

dergraf:
fixed in #271

PR #271, opened 2017-02-10
Fix multiple bridge issue

larshesel:
Fix for #269 

This is build on top of #270 , so that needs to be merged first and this PR rebased on top.


dergraf:
LGTM!

PR #275, opened 2017-02-14
Increase the KeepAlive timeout to 1.5x the connection's keepalive

kellycampbell:
This is per section 3.1.2.10 of the MQTT 3.1.1 spec:

  If the Keep Alive value is non-zero and the Server does not receive a
  Control Packet from the Client within **one and a half times** the
  Keep Alive time period, it MUST disconnect the Network Connection to
  the Client as if the network had failed [MQTT-3.1.2-24].

A client with a 20-second Keepalive was sometimes being disconnected before it
could send its next pingreq:

<0.20829.0>@vmq_mqtt_fsm:connected:411 [<0.20829.0>] stop due to keepalive expired


larshesel:
Hi @kellycampbell thanks for the PR - we'll try to review as soon as we have a moment free.

dcy:
```
get_right_keepalive(KeepAlive) ->
    (KeepAlive * 3) div 2.
```
Maybe this is better, change the value in one place.

larshesel:
Hi all, I've created another PR #294 which should fix this issue.

The PR doesn't change when the keepalive timer is triggered (a heuristic of 0.75 * the client keepalive value), but will disconnect when, in a check, the last incoming packet is longer ago than 1.5 times the client keepalive value. I hope that makes sense.

dergraf:
fixed in #294

kellycampbell:
Sounds great. Thanks!

larshesel:
No problem, sorry for the delay.

PR #289, opened 2017-03-02
Remove default bridge configurations

larshesel:
The user should define this explicitly.

This makes sure only explicitly defined bridges are activated.


dergraf:
LGTM!

PR #291, opened 2017-03-03
Add connack counter

larshesel:
Proposal to add a connack counter as suggested in erlio/vmq_server#117


dergraf:
How about generalize it to have a counter for every MQTT connack return code?

larshesel:
How does this look @dergraf ?

dergraf:
LGTM!

PR #294, opened 2017-03-03
Disconnect earliest at 1.5 times keepalive

larshesel:
Make sure that we will not disconnect a connection unless we're sure at
least 1.5 times the keepalive time has passed.

Also, rename the timer to reflect that it's a timer to check if
something should be done regarding the keepalive value from the client.


dergraf:
LGTM!

PR #309, opened 2017-03-15
VerneMQ session tracing

larshesel:
If you wanna try @ioolkos :)

To get started run `vmq-admin trace` and see where that takes you...


larshesel:
Need to figure how to handle/not handle mountpoints. Also this currently will only trace a single session and just chooses one if there are multiple when starting the trace and when a new session is created that one will be traced.

apps/vmq_server/src/vmq_tracer.erl
```diff
@@ -0,0 +1,472 @@
+-module(vmq_tracer).
```
dergraf: Add Copyright

larshesel: Done

larshesel:
@dergraf @ioolkos I've cleaned up the code and added more documentation (also a bit on the `vmq-admin` side to make it clear the tracing sub-system cannot be considered stable in features and output). But this is my candidate for merging if we want it to ship with 1.0 - I'm sure there are a *lot* of things that could be much better than they are now :)

larshesel:
@drf maybe this would also be of interest to you to play with - it's likely it'll go into the 1.0 release which is imminent.

dergraf:
Looks freaking awesome!

PR #323, opened 2017-03-22
Stop all_till_ok chain on anything but `next`

larshesel:
Make sure to stop `all_till_ok` hook evaluation if a hook returns `ok`, `error` or anything that is not `next`.


apps/vmq_plugin/src/vmq_plugin_helper.erl
```diff
@@ -23,13 +23,13 @@ all([{Module, Fun}|Rest], Params, Acc) ->
...
         {ok, V} -> {ok, V};
-        E -> all_till_ok(Rest, Params, [E|ErrAcc])
+        {error, Error} -> {error, Error};
+        next -> all_till_ok(Rest, Params);
+        E -> {error, E}
     end;
-all_till_ok([], _, ErrAcc) ->
-    {error, ErrAcc}.
+all_till_ok([], _) ->
+    {error, next}.
```
dergraf: maybe we could use `chain_exhausted` instead of `next`, what do you think=

larshesel: I like that

dergraf:
LGTM!

Issue #324, opened 2017-03-22
Refresh CRL files.


larshesel:
### Environment

- VerneMQ Version: any
- OS: any

### Expected behavior

When updating the CRL files the changes should be picked up automatically or there should be an easy way to manually refresh them

### Actual behaviour

CRL file changes are not picked up.

See also #321

larshesel:
closed via #326

PR #326, opened 2017-03-23
Automatically refresh (purge) crls

larshesel:
About the simplest solution I could think of for #324


dergraf:
LGTM!

larshesel:
@dergraf also added the changelog entry - only thing I'm unsure about is if the option should be hidden like it currently is.

PR #327, opened 2017-03-24
Fix missed cli-cleanup change

larshesel:
We now consistently use `show` instead of `status`.


dergraf:
LGTM! Merge due to triviality.

PR #328, opened 2017-03-24
Hide metrics reset command

larshesel:
This command is only meant for debugging purposes and can in some cases
lead to inconsistent counters.


dergraf:
LGTM!

Issue #339, opened 2017-04-01
webhooks: on_unsubscribe crash


funwun:
### Environment

- VerneMQ Version: erlio/docker-vernemq:1.0.0

When I use plugins.vmq_webhooks = on and I register the endpoint of on_unsubscribe, then crash:

ERROR: 
```
2017-04-01 07:17:04.699 [error] <0.2149.0> CRASH REPORT Process <0.2149.0> with 0 neighbours crashed with reason: no case clause matching {ok,[]} in vmq_reg:unsubscribe_op/3 line 105
2017-04-01 07:17:04.699 [error] <0.2149.0> Ranch listener {{172,17,0,2},1883} terminated with reason: {case_clause,{ok,[]}}
```

larshesel:
Thanks for reporting, this seems to be an small oversight on the part of the webhooks plugin, we'll investigate.

larshesel:
Hi @funwun - I've created  fix for this and it'll hopefully be part of the next bugfix release (1.0.1)

funwun:
@larshesel Thanks, I'm looking forward to this release (1.0.1)

Issue #332, opened 2017-03-27
Overriding mountpoint in webhooks has no effect


larshesel:
### Environment

- VerneMQ Version: master
- OS: any
- Erlang/OTP version (if building from source): 19.3

### Expected behavior

When overriding the mountpoint in a webhook I would expect the client to be assigned the new mointpoint

### Actual behaviour

The client still has the old mountpoint.

Note: this has to do with the fact that the `vmq_mqtt_fsm` only expects the modifier `subscriber_id` and not the individual `client_id` and `mountpoint`.

larshesel:
The solution here would be to return a `subscriber_id` for the `auth_on_register` hook from within the webhooks (and diversity plugin) instead of separate `client_id` and `mountpoint` values so it conforms with the documented  (and of `vmq_mqtt_fsm` expected) behaviour of the `auth_on_register` hook.

PR #343, opened 2017-04-03
Convert empty modifier list to just `ok`

larshesel:
Fix for #339


larshesel:
@dergraf hold on reviewing this - I think I'll replace it with on which solves #332 as well

larshesel:
Replaced by #345

PR #345, opened 2017-04-03
Handle subscriber id modifier

larshesel:
This moves the `check_modifier/1` into `vmq_plugin` so it can be reused from `vmq_diversity` and `vmq_wehbooks`.

This solves both #339 and #332 

This PR replaces #343

### DIFF
--- a/apps/vmq_plugin/src/vmq_plugin_util.erl
+++ b/apps/vmq_plugin/src/vmq_plugin_util.erl
@@ -1,0 +1,142 @@
+%% Copyright 2017 Erlio GmbH Basel Switzerland (http://erl.io)
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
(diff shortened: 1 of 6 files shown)


apps/vmq_plugin/src/vmq_plugin_util.erl
```diff
@@ -0,0 +1,129 @@
...
+
+-export([check_modifiers/2]).
+
+check_modifiers(auth_on_subscribe, Modifiers) ->
+    %% auth_on_subscribe take a special form
+    lists:foldl(fun (_, error) -> error;
+                    ([T, Q], Acc) when is_binary(T) and is_number(Q) ->
+                        case vmq_topic:validate_topic(subscribe, T) of
+                            {ok, Topic} ->
+                                [{Topic, to_internal_qos(round(Q))}|Acc];
```
larshesel: @dergraf the `to_internal_qos/1` function is the only change from the original `check_modifiers/2` function.

dergraf: not sure if I like this module to be part of vmq_plugin, but besides that it LGTM.

larshesel: Yeah, also thought of that - the only other place to put it at the moment would be in vmq_commons or vernemq_dev. But then on the other hand it's kind of strongly related to plugins...

larshesel: As I don't think we came up with a better place for the `check_modifiers/2` function yet I guess we'll leave it in `vmq_plugin_utils.erl` for now.

I've rebased, squashed and updated the changelog. Ok to merge if tests pass?

dergraf: Yed

larshesel: I'll take that as a misspelled 'Yes'.

PR #346, opened 2017-04-03
Fix vmq_admin session show

dergraf:
This should fix the combinator issues we've seen with `vmq-admin session`.

I'll add more testing

### DIFF
--- a/apps/vmq_ql/.gitignore
+++ b/apps/vmq_ql/.gitignore
@@ -1,0 +1,1 @@
+src/vmq_ql_parser.erl
--- a/apps/vmq_ql/include/vmq_ql.hrl
+++ b/apps/vmq_ql/include/vmq_ql.hrl
@@ -1,6 +1,6 @@
 -record(vmq_ql_table, {
           name                  :: atom(),
-          depends_on            :: [#vmq_ql_table{}],
+          depends_on = []       :: [#vmq_ql_table{}],
           provides = []         :: [atom()],
           init_fun              :: function(),
           include_if_all = true :: boolean()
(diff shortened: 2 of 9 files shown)


larshesel:
Playing with it I found this example which hangs - I have no idea at all if this is supposed to work or not:

```
(VerneMQ@127.0.0.1)3> Q = "SELECT * FROM sessions WHERE client_id LIKE 'cli%'".            
"SELECT * FROM sessions WHERE client_id LIKE 'cli%'"
(VerneMQ@127.0.0.1)4> vmq_ql_query_mgr:fold_query(fun(Row, Acc) -> [Row, Acc] end, [], Q). 
```

I had one offline client at that point:
```
(VerneMQ@127.0.0.1)1> Q5 = "SELECT * FROM sessions WHERE  is_online=false".                                                   
"SELECT * FROM sessions WHERE  is_online=false"
(VerneMQ@127.0.0.1)2> vmq_ql_query_mgr:fold_query(fun(Row, Acc) -> [Row, Acc] end, [], Q5).
[#{'__mountpoint' => [],clean_session => false,
   client_id => <<"client">>,deliver_mode => fanout,
   is_offline => true,is_online => false,is_plugin => false,
   mountpoint => <<>>,node => 'VerneMQ@127.0.0.1',
   num_sessions => 0,offline_messages => 1,
   online_messages => 0,queue_pid => <0.317.0>,
   statename => offline},
 []]
```

dergraf:
Oh damn, have to cleanup the deprecated `random` stuff in the test module.

larshesel:
LGTM!

PR #347, opened 2017-04-04
Purge stale subscriber data when (re)starting VerneMQ

larshesel:
When a VerneMQ node is stopped, the clean session true session data
which is persisted to disk will not neccesarily be deleted and will
still be there when VerneMQ starts.

This commit will make sure to delete all node-local subscriptions which
are made by a clean session true subscriber when starting VerneMQ.

### DIFF
--- a/apps/vmq_server/test/vmq_cluster_SUITE.erl
+++ b/apps/vmq_server/test/vmq_cluster_SUITE.erl
@@ -18,7 +18,8 @@
          shared_subs_random_policy_test/1,
          shared_subs_prefer_local_policy_test/1,
          shared_subs_local_only_policy_test/1,
-         cross_node_publish_subscribe/1]).
+         cross_node_publish_subscribe/1,
+         restarted_node_has_no_stale_sessions/1]).
 
 -export([hook_uname_password_success/5,
          hook_auth_on_publish/6,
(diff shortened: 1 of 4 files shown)


larshesel:
Yes! We definitely should - I'll write some up.

larshesel:
@dergraf added a test for this - was unsure how to test this kind of thing so ended up with this... let me know what you think...

apps/vmq_server/test/vmq_cluster_SUITE.erl
```diff
@@ -543,6 +545,56 @@ shared_subs_random_policy_test(Config) ->
...
+          fun(#{node := N, client_id := C} = S, {Removed, Other})
+                when N =:= RestartNodeName, C =:= ClientId ->
+                  {[S|Removed], Other};
+             (S, {Removed, Other}) ->
+                  {Removed, [S|Other]}
+          end,
+          {[], []},
+          Sessions),
+
+    [] = Removed,
```
dergraf: Not sure about this Removed check, but otherwise LGTM!

larshesel: Hey @dergraf I rewrote the test to be more directly comparing things... I think (hope) it's clearer now.

dergraf:
Not sure if `vmq_reg` should be the module holding this function (or we may add `maybe_change_nodename` to `vmq_reg` too). Otherwise this LGTM!

larshesel:
Move the purge_stale... function to vmq_reg. Ok to Merge if tests pass?

PR #349, opened 2017-04-06
Fix cross node multiple subscriber issue

larshesel:
I'll add some tests for this case as well so we don't experience this in the future.


dergraf:
Good catch, LGTM!

larshesel:
@dergraf I've added a test and did a small refactor (renaming of some helper functions to be more general). Can you take quick look  before I squash/rebase/merge  (if tests pass, of course)?

PR #351, opened 2017-04-07
handle dynamic cowboy routes

dcy:
Hi!
When ```vmq-admin plugin enable xxx_plugin```, the routes of xxx_plugin will work.
And when ```vmq-admin plugin disable xxx_plugin```, the routes of xxx_plugin will not work.
It is dynamic if i add ```{http_modules, [xxx_mod]}``` to ```xxx_plugin.app.src``` or ```advanced.config````


dergraf:
Hello dcy,

thank you for your contribution. I like the idea, but don't think we should solve it the way you have proposed. The `vmq_plugin` app internals shouldn't call directly into `vmq_server` application internals. We have to come up with a different solution for this issue. In any case, thank you very much for your contribution, highly appreciated.

Best!
Andre

PR #352, opened 2017-04-08
Report ets table metrics in bytes instead of wordsize

larshesel:
To me it's a bug that we report the sizes of ets tables (retained and subscription data) in wordsizes which vary depending on the platform (4bytes on 32bit and 8 bytes on 64 bit platforms), so I changed these metrics so they're reported in bytes instead.


dergraf:
Good point. If you are reviewing those metrics. Can you double check that all involved ETS tables are counted?

larshesel:
I added the size of the retain update ets table to the reported memory usage for retained messages. All tables in `vmq_reg_trie` are accounted for.

dergraf:
LGTM!

PR #353, opened 2017-04-09
Add descriptions to metrics

larshesel:
I still need to get the cli usage right.

Let me know what you think (and especially if you have corrections to the descriptions!).

Since this is new functionality I'm putting this into the 1.1.0 milestone.

### DIFF
--- a/apps/vmq_server/src/vmq_metrics.erl
+++ b/apps/vmq_server/src/vmq_metrics.erl
@@ -72,6 +72,7 @@
         ]).
 
 -export([metrics/0,
+         metrics/1,
          check_rate/2,
          reset_counters/0,
          reset_counter/1,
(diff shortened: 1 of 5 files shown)


larshesel:
I thought about that, but then stopped thinking about it because I couldn't decide for one or the other. Not sure what you mean by 'in a general way'.

larshesel:
Thinking a bit about it, it would probably make sense to have the descriptions right next to the actual metrics - then we can extract everything for the cli tools in one go if we need it - I suppose that's what you meant by `a general way` ;)

larshesel:
Moved the metric descriptions into the vmq_metric module and added an `metrics/1` function which, if passed true, retrurns metrics with descriptions.

dergraf:
LGTM!

larshesel:
Hi @dergraf. Took a look at your changes and it looks good. Did you test the helplines with prometheus? There's no chance that'll break something somehow?

I've changed this to point so it'll be merged into the `dev-1.1.0` branch instead of master.

Only question I have is should we do something about the 'deprecated' counter: https://github.com/erlio/vernemq/pull/353/files#diff-f6400fa1f04176af5d2971f0f916c87bR573 in this PR or fix that as a bugfix later?

larshesel:
Also rebased the branch on top of `dev-1.1.0` and squashed to a single commit.

PR #356, opened 2017-04-11
Merge backup queue with the session queue during session teardown

dergraf:
@larshesel please review

### DIFF
--- a/apps/vmq_server/src/vmq_queue.erl
+++ b/apps/vmq_server/src/vmq_queue.erl
@@ -722,10 +722,10 @@ insert_from_session(#session{queue=Queue},
     %% allow other sessions to balance the messages of the dead queue
     insert_from_queue(Queue, State).
 
-insert_from_queue(#queue{type=fifo, queue=Q}, State) ->
-    insert_from_queue(fun queue:out/1, queue:out(Q), State);
-insert_from_queue(#queue{type=lifo, queue=Q}, State) ->
-    insert_from_queue(fun queue:out_r/1, queue:out_r(Q), State).
+insert_from_queue(#queue{type=fifo, queue=Q, backup=BQ}, State) ->
+    insert_from_queue(fun queue:out/1, queue:out(queue:join(BQ, Q)), State);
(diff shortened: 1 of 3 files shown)


apps/vmq_server/src/vmq_queue.erl
```diff
@@ -815,9 +815,10 @@ send_notification(#session{pid=Pid} = Session) ->
     vmq_mqtt_fsm:send(Pid, {mail, self(), new_data}),
     Session#session{status=passive}.
 
-cleanup_session(SubscriberId, #session{queue=#queue{queue=Q}}) ->
+cleanup_session(SubscriberId, #session{queue=#queue{queue=Q, backup=BQ}}) ->
     _ = vmq_metrics:incr_queue_unhandled(queue:len(Q)),
```
dergraf: Should we include the size of a potentially filled backup queue in the 'unhandled' counter? At this point we actually don't know if the fsm has flushed those messages or not.

larshesel: Good question. I'd say we shouldn't as it would be guesswork on the part of the queue and I suppose those messages would only be unhandled if the fsm crashes and doesn't do the right thing?

larshesel:
This definitely fixes the issue - is this a case where we'd want a test?

changelog.md
```diff
@@ -15,6 +15,8 @@
 - Fix formatting bug in the 'vmq-admin trace` command.
 - Handle empty modifier list correctly in `vmq_webhooks` (#339).
 - Handle client_id and mountpoint modifiers correctly in `vmq_webhooks` (#332).
+- Fix bug where an aborted connection handshake caused queued offline messages 
+  not being migrated to remote node.
```
larshesel: I'd rather say `migrated properly` as the messages are migrated to the remote node, they're just not available in the queue.

larshesel:
Looks great - don't forget to edit the message in the merge commit after squashing..

dergraf:
@larshesel approve?

Issue #344, opened 2017-04-03
FreeBSD: vernemq failed to start within 15 seconds


hieutmd:
### Environment

- VerneMQ Version: 1.0
- OS: FreeBSD 11.0
- Erlang/OTP version (if building from source): 19.2

### Expected behavior

- Should start after successful install

### Actual behaviour

`
vernemq failed to start within 15 seconds,
see the output of 'vernemq console' for more information.
If you want to wait longer, set the environment variable
WAIT_FOR_ERLANG to the number of seconds to wait.
`

When run vernemq console, get:
`
=INFO REPORT==== 4-Apr-2017::01:36:19 ===
    alarm_handler: {set,{{disk_almost_full,"/dev"},[]}}
[os_mon] cpu supervisor port (cpu_sup): Erlang has closed
{"Kernel pid terminated",application_controller,"{application_start_failure,plumtree,{{error,{shutdown,{failed_to_start_child,plumtree_metadata_leveldb_instance_sup,{shutdown,{failed_to_start_child,plmtrlvld_0,{undef,[{eleveldb,validate_options,[open,[{block_cache_threshold,33554432},{block_restart_interval,16},{block_size_steps,16},{compression,true},{create_if_missing,true},{data,[{dir,\"./data\"}]},{data_root,\"./data/leveldb\"},{delete_threshold,1000},{eleveldb_threads,71},{fadvise_willneed,false},{included_applications,[]},{limited_developer_mem,false},{sst_block_size,4096},{sync,false},{tiered_slow_level,0},{total_leveldb_mem_percent,6},{use_bloomfilter,true},{verify_checksums,true},{verify_compaction,true},{write_buffer_size,45659641},{write_buffer_size_max,62914560},{write_buffer_size_min,31457280}]],[]},{plumtree_metadata_leveldb_instance,init_state,2,[{file,\"/usr/home/hieu/vernemq/_build/default/lib/plumtree/src/plumtree_metadata_leveldb_instance.erl\"},{line,299}]},{plumtree_metadata_leveldb_instance,init,1,[{file,\"/usr/home/hieu/vernemq/_build/default/lib/plumtree/src/plumtree_metadata_leveldb_instance.erl\"},{line,131}]},{gen_server,init_it,6,[{file,\"gen_server.erl\"},{line,328}]},{proc_lib,init_p_do_apply,3,[{file,\"proc_lib.erl\"},{line,247}]}]}}}}}},{plumtree_app,start,[normal,[]]}}}"}
[...]

Issue #336, opened 2017-03-31
make rel error on FreeBSD 10.2 


hieutmd:
### Environment

- VerneMQ Version: 1.0.0
- OS: FreeBSD 10.2 
- Erlang/OTP version (if building from source): 19.3

### Expected behavior

Build to be successfull

### Actual behaviour

Run gmake sets following errors:

`gmake rel`

`gmake[1]: Leaving directory '/usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src/leveldb'
make[1]: "/usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src/Makefile" line 19: Missing dependency operator
make[1]: "/usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src/Makefile" line 24: Missing dependency operator
make[1]: "/usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src/Makefile" line 28: Missing dependency operator
make[1]: "/usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src/Makefile" line 32: Need an operator
make[1]: "/usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src/Makefile" line 65: warning: duplicate script for target "%.o" ignored
make[1]: "/usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src/Makefile" line 62: warning: using previous script for "%.o" defined here
make[1]: "/usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src/Makefile" line 68: warning: duplicate script for target "%.o" ignored
make[1]: "/usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src/Makefile" line 62: warning: using previous script for "%.o" defined here
make[1]: "/usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src/Makefile" line 71: warning: duplicate script for target "%.o" ignored
make[1]: "/usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src/Makefile" line 62: warning: using previous script for "%.o" defined here
make[1]: Fatal errors encountered -- cannot continue
make[1]: stopped in /usr/home/userx/vernemq/_build/default/lib/eleveldb/c_src
===> Hook for compile failed!

gmake: *** [Makefile:56: rel] Error 1
`

hieutmd:
Please advise on the issue.

iblis17:
`cd /path/to/vernemq/_build/default/lib/eleveldb`
[...]

PR #362, opened 2017-04-13
vmq_passwd: fix build on freebsd

iblis17:
See also: #344, #336

### DIFF
--- a/apps/vmq_passwd/c_src/Makefile
+++ b/apps/vmq_passwd/c_src/Makefile
@@ -1,3 +1,4 @@
+CC?=	gcc
 
 compile:
-	gcc $(CFLAGS) vmq_passwd.c -lcrypto -o ../priv/vmq_passwd
+	$(CC) $(CFLAGS) vmq_passwd.c -lcrypto -o ../priv/vmq_passwd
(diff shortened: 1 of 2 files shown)


larshesel:
Hi @iblis17 I think this looks good - I'd however like to have all changes needed to compile VerneMQ on freebsd on a single branch so we can test once that it works and that it doesn't affect our builds and package generations for other systems.

I've started a branch over here: https://github.com/larshesel/vernemq/tree/freebsd but it's very incomplete (mostly the work is to get a freebsd-11 system up and running with erlang inside vagrant) and I don't think I'll have too much time the next weeks to move it forward. Let's see.

In any case, let's keep this PR open and include it into the final branch somehow.

iblis17:
Just add some notes.

- [ ] `eleveldb.so` corruption: https://github.com/erlio/vernemq/issues/344#issuecomment-294123113
    - workaround: https://github.com/erlio/vernemq/issues/344#issuecomment-293807893
- [ ] erlio/eleveldb#1

iblis17:
@dergraf I am on a tight schedule this week... I will catch it up soon.

iblis17:
👌

PR #375, opened 2017-04-28
log keepalive expire detail

dcy:
Hi!
Want to know which client expire


larshesel:
I think it looks ok as well, one grammatic nitpic is that it shoud be `stopped` instead of `stop`, so the line should read: `client ~p with username ~p stopped due to keepalive expired`.

larshesel:
I think this should also be mentioned in the changelog. Something like a line saying `Add more detail to the client expiration log message`. I guess it would be helpful if we added some kind of contribution guideline/help document for future contributors :)

larshesel:
OK for me - @dergraf ?

larshesel:
Thanks @dcy !

PR #381, opened 2017-05-13
sandbox Lua calls

dergraf:


larshesel:
Looks good to me, but an entry in the changelog should be added.

ioolkos:
vmq_in_order_delivery_SUITE fails on OTP 19.3 (travis-ci)

larshesel:
That's something which pops up from time to time, it's likely a race-condition in the test. I restarted the build for OPT 19.3.

PR #387, opened 2017-05-17
Intercept FSM messages in terminated state

dergraf:


apps/vmq_server/src/vmq_websocket.erl
```diff
@@ -70,6 +70,9 @@ websocket_info({set_sock_opts, Opts}, Req, State) ->
     [Socket, Transport] = cowboy_req:get([socket, transport], Req),
     Transport:setopts(Socket, Opts),
     {ok, Req, State};
+websocket_info({FsmMod, _}, Req, #st{fsm_mod=FsmMod, fsm_state=terminated} = State) ->
+    % we got an intermediate message before retrieving {?MODULE, terminate}
+    {shutdown, Req, State};
```
larshesel: Could you add a comment in the commit or here in the code which explains it's OK to ignore these messages (while in the terminated state). What kind of messages is it we're throwing away here?

Further, could you make the comment a `%%` instead of a single `%`. We use two everywhere else (and the emacs-mode will indent a single one to the far right margin... O_o

larshesel:
A changelog item should probably also be added (maybe just something like 'general bugfixes')

PR #397, opened 2017-05-24
Forward connections without username/password to next hook in chain

baranga:
`vmq_passwd` rejected connections without username/password with `error` which prevents further handling in later hooks in chain. This is now changed to forward the event to next hook by returning `next` for connections without username/password.

This should fix erlio/vernemq#396.

### DIFF
--- a/apps/vmq_passwd/src/vmq_passwd.erl
+++ b/apps/vmq_passwd/src/vmq_passwd.erl
@@ -99,9 +99,9 @@ load_from_list(List) ->
     del_aged_entries().
 
 check(undefined, _) ->
-    {error, not_authorized};
+    next;
 check(_, undefined) ->
-    {error, invalid_credentials};
+    next;
 check(User, Password) ->
     case ets:lookup(?TABLE, ensure_binary(User)) of
         [{_, SaltB64, EncPassword, _}] ->
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@
 - Add more detail to the client expiration log message.
 - Safeguard Lua function calls
 - Fix bug where websocket connections were incorrectly terminated (#387).
+- Fix breakage of event hook chain in `vmq_passwd` (#396).
 
 ## VERNEMQ 1.0.1
 


larshesel:
Changing this might trigger unexpected behaviour in plugins later in the chain and be a security problem if they relied on this behaviour.

I guess this is also a breaking change - so we wouldn't be able to change this until 2.0.0.

An alternative would be to make this behaviour configurable so it by default is backwards compatible but still able to solve your problem.

Just me 2 cents - @dergraf what do you think?

ioolkos:
What is actually more logical and user friendly? I think it's the way proposed in this fix.

Is there really a case where someone could've relied on this? trying to wrap my head around it :)

baranga:
Scenarios I can think of:
- `vmq_passwd` not activated, no impact
- `vmq_passwd` is activated
  - but is the only plugin providing  `auth_on_register` hook, so connection handling isn't changed because connection gets dropped after last hook anyway (`allow_anonymous = off`, only error message is different)
  - provides the first `auth_on_register` hook, so before change other hooks never get called (on missing username/password) but might get called now
  - provides last `auth_on_register` hook, same as if it's the only one

So the "breaking" change would be plugins get called without username/password. I would assume current implementations to not handle this and return `next` or `error` in this cases which would also not change the overall outcome.

But yes, this could be confusing for operators. The config option might be the best.

ioolkos:
I'm of the opinion to just change it and advertise it as 'breaking' (or half breaking or whatever).

I also would prefer not to introduce a config option that is rather opaque and will have to stay there forever.

baranga:
@dergraf Note should go to `1.0.2` subsection, right?

ioolkos:
jep, tag 1.0.2.
Thanks!

larshesel:
@dergraf as it's considered a bug I'm absolutely :+1: on this if the proper changelog entry is added.

baranga:
Updated `changelog.md`

larshesel:
Hi @baranga - your PR just got merged :)

PR #404, opened 2017-06-04
Load balance Lua scripts

### LINKED ISSUE #382: Remove Lua script bottleneck
Feature Request:

Currently a Lua script and it's Lua state is kept in a gen_server per script. As a consequence a call to the same hook in the script required to go through the same gen_server. 
Therefore, the built-in connection pools for the various database backends don't help to distribute load as long as only a single script is used.
The single-script/single-gen_server case is the approach which is currently in use for all the provided auth_cache solutions.

Solution for this problem:

Assuming no Lua state manipulations happens during the execution of a script (as it is the case for the auth_cache) we could have multiple gen_servers that serve the same script and load balance the Lua function calls. The `keep_state` config variable (https://github.com/erlio/vernemq/blob/master/apps/vmq_diversity/src/vmq_diversity_script.erl#L82) could indicate whether the script has side-effects wrt. to the Lua state. If it hasn't side-effects we could implement a transparent load balancing.

What do you think @larshesel / @ioolkos

dergraf:
fixes #382

### DIFF
--- a/apps/vmq_diversity/priv/auth/auth_commons.lua
+++ b/apps/vmq_diversity/priv/auth/auth_commons.lua
@@ -63,3 +63,6 @@ end
 function on_client_offline(c)
 end
 
+-- set for every DB auth handler
+num_states = 10
+keep_state = false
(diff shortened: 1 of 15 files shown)


dergraf:
Currently missing:
Testing of scheduling logic in `vmq_diversity_script`
Add tests for reloading scripts
Add changelog entry

PR #425, opened 2017-06-25
Don't build on Erlang 19.2

larshesel:
Let's get rid of the 19.2 build (we're building 19.3 anyway and we should switch our docker images to it as well).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ language: erlang
 install: true
 otp_release:
     - 18.3
-    - 19.2
     - 19.3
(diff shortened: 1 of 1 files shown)


codeadict:
what about 20 :)

larshesel:
20 needs some changes to some of the dependencies - and to VerneMQ itself. I did a quick test and got it running within 10 minutes, so it's not really complicated and we'll try to get it running out of the box on 20 soon. PRs appreciated :D

larshesel:
@codeadict also see #426

Issue #402, opened 2017-05-30
Misunderstanding: `message_size_limit` and `max_message_size`


dergraf:
Keep: `max_message_size` to limit TCP parser buffer
New: `max_payload_size` to limit the MQTT payload size,

Alias: `message_size_limit => max_payload_size`

Deprecate `message_size_limit` in 1.1.0

larshesel:
Looking closer at the code I don't see we have anything else than the max size passed to the `vmq_parser:parse/2` function which will implicitly limit the payload size. I don't see that we have anything limiting a TCP parser buffer.

I would therefore just keep `max_message_size` (which will then be the only one passed to `vmq_parser:parse/2`) and deprecate `message_size_limit`.

Maybe I'm missing something ;)

dergraf:
Yes, you're right! We should keep `max_message_size`.

PR #428, opened 2017-06-27
Fix `message_size_limit` / `max_message_size` bug

larshesel:
Only one option should ever have existed. `message_size_limit` has now
been deprecated and it has no effect if used. Use `max_message_size`
instead.

### DIFF
--- a/apps/vmq_server/src/vmq_config_cli.erl
+++ b/apps/vmq_server/src/vmq_config_cli.erl
@@ -41,7 +41,7 @@ register_config_() ->
      "queue_deliver_mode",
      "queue_type",
      "max_message_rate",
-     "message_size_limit",
+     "max_message_size",
      "upgrade_outgoing_qos",
      "systree_enabled",
      "systree_interval",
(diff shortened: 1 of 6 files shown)


larshesel:
Reference #403  #402

apps/vmq_server/src/vmq_config_cli.erl
```diff
@@ -41,7 +41,6 @@ register_config_() ->
      "queue_deliver_mode",
      "queue_type",
      "max_message_rate",
-     "message_size_limit",
```
dergraf: should we add max_message_size here?

apps/vmq_server/src/vmq_server.app.src
```diff
@@ -30,7 +30,6 @@
       {retry_interval, 20},
       {max_inflight_messages, 20},
       {max_message_rate, 0}, % no rate limit
-      {message_size_limit, 0},
```
dergraf: Should we add max_message_size here?

apps/vmq_server/src/vmq_config_cli.erl
larshesel: Added

apps/vmq_server/src/vmq_server.app.src
larshesel: Added

PR #429, opened 2017-06-28
Forget the old plugin code path

larshesel:
With the current implementation we add plugin code paths to the end of the code path and this is state that we manage. If we do not remove it and upgrade a plugin the code path of the new plugin will end up at the end of the code path and calls like `code:lib_dir/1` and `code:priv_dir/1` will return the path of the old plugin.

### DIFF
--- a/apps/vmq_plugin/src/vmq_plugin_mgr.erl
+++ b/apps/vmq_plugin/src/vmq_plugin_mgr.erl
@@ -545,7 +545,7 @@ check_app_plugin(App, Options) ->
             lager:debug("can't create paths ~p for app ~p", [AppPaths, App]),
             {error, plugin_not_found};
         Paths ->
-            code:add_paths(Paths),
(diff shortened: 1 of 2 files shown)


larshesel:
Seems to break a bunch of tests - likely because we're trying to disable some plugins that aren't enabled.

larshesel:
It didn't work because if I remove the code path of a plugin this breaks the internal plugins if they are disabled and enabled - on the second enable the plugin application is no longer in the code path.

So the workaround is to add the plugin code paths to the beginning of the list of code paths. This doesn't fix the problem that plugins with different versions that are upgraded will leave old code paths behind, but we already had  that problem.

Only concern I have is this: will it have some negative effects that plugins now come first in the code path? Could they shadow for things they shouldn't (if they have dependencies, for instance...)

larshesel:
@dergraf can I request a second review for this (no hurry)

PR #431, opened 2017-06-29
VMQ-426 support for Erlang 20

codeadict:
This is still **WIP**


apps/vmq_commons/.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: erlang
 otp_release:
-    - 17.5
-    - 18.3
+    - 10.3
```
gleber: Should be 19.3

codeadict: ahh thanks, typo :)

larshesel:
I've been thinking a bit more this and I think the better way to go would be to not replace gen_fsm with gen_statem just yet. In the long run we'd need to move there, but the general policy is to keep deprecated functionality around for two major releases, so there's no rush to move to gen_statem. So I think the easy and low risk path to get onto OTP20  would be to add some `{nowarn_deprecated_function, MFAs}` to the compile attribute in the relevant modules.

codeadict:
Ahh yeah that sounds way easier, will do that for now, same for Plumtree and we will be ready to roll

dergraf:
closing in favor of #450 . Thanks a ton for your efforts!

PR #450, opened 2017-07-19
WIP: Erlang/OTP 20 support

larshesel:
Branch where we can merge all the OTP20 changes until we have a complete working PR.


larshesel:
@dergraf I squashed everything into a single commit. I think we can merge this as I do not think it would have any negative effects on Erlang/OTP 19 builds and we can open up for people to start working with OTP20.

larshesel:
Only outstanding issue AFAIK is we're using my fork of cuttlefish until my PR gets merged...

Issue #475, opened 2017-08-04
VerneMQ does not reconnect to mongoDB if Database is restarted


KnolBee:
### Environment

- VerneMQ Version: 1.1.0
- OS: Ubuntu 16.04
- Erlang/OTP version (if building from source):

### Expected behavior

- Restart of MongoDB shouldn't force restart of VerneMQ

### Actual behaviour

- If MongoDB is restarted, VerneMQ stops authenticating until restarted.

ioolkos:
@KnolBee thanks for pointing that out. That's totally unfortunate.
@larshesel @dergraf Is there some trick to restart the connection pool? or should it work out of the box anyway?

larshesel:
This looks like a bug - I'll investigate soon.

larshesel:
I can confirm this - the child connections are restarted until the supervisor reaches its max restart intensity and shuts down taking vmq_diversity with it.

KnolBee:
Thanks @larshesel are there plans to fix this? This is definitely a blocker for production use of MongoDB authentication module.

larshesel:
It will be fixed but I can't give a time frame ATM.

larshesel:
Btw. I forgot to answer if there was some workaround to this. I would expect - but haven't tested it - that if you restart the `vmq_diversity` plugin then the pools should get reinitialized. I'll try to spend some time on this issue today.

larshesel:
Hi @KnolBee , I've a branch in #491 which should fix this issue. Could you test that out? Next I'd try to bring in an upgraded mongodb driver.

larshesel:
Merged

KnolBee:
Hi @larshesel Ok great, Will test over the weekend.

PR #491, opened 2017-08-24
Use wrapper for mongo client to protect poolboy

larshesel:
Use wrapper for mongo client to protect poolboy if the mongodb database
is not available.

I think we'd need to do the same for the postgres and maybe other drivers, but I thought I'd get the approach reviewed first. The wrapper I wrote I tried to keep general with the hope we can use it for other cases as well.

This is a first step toward fixing #475

### DIFF
--- a/apps/vmq_diversity/src/vmq_diversity_worker_wrapper.erl
+++ b/apps/vmq_diversity/src/vmq_diversity_worker_wrapper.erl
@@ -1,0 +1,221 @@
+%% Copyright 2017 Erlio GmbH Basel Switzerland (http://erl.io)
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+
+%% @doc Simple gen_server for poolboy used to wrap a client process
+%% and restart (after a configurable timeout) it if it dies for some
+%% reason. The purpose is to protect poolboy from dying if an external
+%% resource is not available and all the client processes die.
+%% @end.
+-module(vmq_diversity_worker_wrapper).
+
+-behaviour(poolboy_worker).
+-behaviour(gen_server).
+
+%% API
+-export([start_link/1, apply/4]).
+
+%% gen_server callbacks
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2,
+         terminate/2, code_change/3]).
+
+-define(SERVER, ?MODULE).
+
+-record(state, {
+          %% Name used for logging purposes.
+          name = undefined,
+
+          %% fun of arity 0 to start the worker.
+          start_fun = undefined,
+
+          %% fun of arity 1 used to terminate the worker when shutting
+          %% down.
+          terminate_fun = undefined,
+
+          %% The worker PID if running.
+          worker = undefined :: {reference(), pid()},
+
+          %% Is the client process ready?
+          connected = false :: boolean(),
+          
+          %% Reconnect timeout if the connection is lost (the worker
+          %% process dies).
+          reconnect_timeout
+         }).
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+%%--------------------------------------------------------------------
+%% @doc
+%% Starts the server
+%%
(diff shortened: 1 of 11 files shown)


larshesel:
Oh, I still need to figure out how the surrounding code handles the `{error, not_connected}` which can now be returned instead of a result...

apps/vmq_diversity/src/vmq_diversity_worker_wrapper.erl
```diff
@@ -0,0 +1,191 @@
...
+
+%% gen_server callbacks
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2,
+         terminate/2, code_change/3]).
+
+-define(SERVER, ?MODULE).
+
+-record(state, {
+          %% module, fun, args used to start the worker.
+          start_mfa = undefined,
```
dergraf: why so explicit?

apps/vmq_diversity/src/vmq_diversity_worker_wrapper.erl
```diff
@@ -0,0 +1,191 @@
...
+%% Handling all non call/cast messages
+%%
+%% @spec handle_info(Info, State) -> {noreply, State} |
+%%                                   {noreply, State, Timeout} |
+%%                                   {stop, Reason, State}
+%% @end
+%%--------------------------------------------------------------------
+handle_info(connect, #state{start_mfa={M,F,A}, worker=undefined} = State) ->
+    NewState =
+    case apply(M,F,[A]) of
```
dergraf: I guess we could force that the provided `A` is always a list. So we have real `apply/3` semantics.

apps/vmq_diversity/src/vmq_diversity_worker_wrapper.erl
```diff
@@ -0,0 +1,191 @@
...
+        {ok, Pid} ->
+            MRef = monitor(process, Pid),
+            State#state{worker={MRef, Pid}};
+        {error, Reason} ->
+            lager:warning("Could not connect due to ~p", [Reason]),
```
dergraf: Let's add some log information containing the `M` and maybe `F`... so one knows which underlying connection is crashing

apps/vmq_diversity/src/vmq_diversity_worker_wrapper.erl
```diff
@@ -0,0 +1,191 @@
...
+%% necessary cleaning up. When it returns, the gen_server terminates
+%% with Reason. The return value is ignored.
+%%
+%% @spec terminate(Reason, State) -> void()
+%% @end
+%%--------------------------------------------------------------------
+terminate(_Reason, #state{worker=undefined}) ->
+    ok;
+terminate(_Reason, #state{worker={Mref, Pid}}) ->
+    mc_worker:disconnect(Pid),
```
dergraf: Hmm, shouldn't this be implementation agnostic.. But for just MongoDB it's ok.

apps/vmq_diversity/src/vmq_diversity_worker_wrapper.erl
larshesel: My idea was to be general and not make any assumptions about the names and arguments.

apps/vmq_diversity/src/vmq_diversity_worker_wrapper.erl
larshesel: Yeah, that's just silly. Hmmm... I guess we'd need to pass in a termination function as well so we can be sure to clean up and not leave any leaked processes behind.

apps/vmq_diversity/src/vmq_diversity_worker_wrapper.erl
larshesel: I'll change the start_mfa to pass in a fun/0 instead and then also add a name we can use here for logging

apps/vmq_diversity/src/vmq_diversity_worker_wrapper.erl
larshesel: I guess it could just be a fun of arity 0 being passed, so we only store a reference to the function, and we wouldn't have to use apply at all for starting the wrapped process. Think that's cleaner.

apps/vmq_diversity/src/vmq_diversity_worker_wrapper.erl
larshesel: Was changed to just call the passed function directly

larshesel:
@dergraf ready for another round of review...

larshesel:
We should do this for the clients that need it - I'll try and get this done this week.

larshesel:
- [ ] mongodb
- [ ] mysql
- [ ] redis
- [ ] postgres
- [ ] memcached

larshesel:
Wrapped the postgresql driver. When postgres isn't there we get a log of crash logs which isn't pretty, but we can't really work around that as that happens within the driver.

larshesel:
I'll rebase after review approval @dergraf

PR #494, opened 2017-08-30
log client disconnected by same client_id

dcy:
Hi!
I think it should log the situation when the client disconnected by same client_id.

### DIFF
--- a/apps/vmq_server/src/vmq_queue.erl
+++ b/apps/vmq_server/src/vmq_queue.erl
@@ -221,6 +221,8 @@ online({add_session, SessionPid, #{allow_multiple_sessions := false} = Opts}, Fr
     %% we've to disconnect currently attached sessions
     %% and wait with the reply until all the sessions
     %% have been disconnected
+    #state{id = SubscriberId} = State,
+    lager:warning("SubscriberId ~p disconnected by same client_id", [SubscriberId]),
     disconnect_sessions(State),
     {next_state, state_change(add_session, online, wait_for_offline),
      State#state{waiting_call={add_session, SessionPid, Opts, From}}};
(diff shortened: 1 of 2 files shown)


apps/vmq_server/src/vmq_queue.erl
```diff
@@ -221,6 +221,8 @@ online({add_session, SessionPid, #{allow_multiple_sessions := false} = Opts}, Fr
     %% we've to disconnect currently attached sessions
     %% and wait with the reply until all the sessions
     %% have been disconnected
+    #state{id = SubscriberId} = State,
+    lager:warning("SubscriberId ~p disconnected by same client_id", [SubscriberId]),
```
larshesel: This should be either a `info` or a `debug` as being disconnected is not an exceptional event but a part of a normal flow. I personally would say it should be on the `debug` level.

larshesel: To conform with other messages it should start with a lower case letter and be on the form we generally use "<action> due to <reason>", so something like: `lager:debug("client ~p disconnected due to multiple sessions not allowed", [SubscriberId])` or something like that, though I'm not 100% sure if the reason I just gave is the most descriptive.

changelog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Not yet release
```
larshesel: Missing `d` at the end of `released`

changelog.md
```diff
@@ -1,5 +1,9 @@
...
+
+- Log client disconnected by same client_id
```
larshesel: I think it would be a bit clearer if it said something like : `- Log when a client is disconnected and multiple sessions are not allowed`.

dergraf:
Just wanted to point out a different angle.. I completely understand the usage of having a log here... What I don't like about logging it with a log level > DEBUG is that two conforming MQTT clients would flood the log stream, as most of the clients will reconnect right away. Opinions? @dcy @larshesel

dcy:
@dergraf We should take some measures(like udid different, redirect to login page) to avoid two clients keep disconnecting another in the development. So it would not flood the log.

larshesel:
Yes, I agree, it should really be a debug level log.

larshesel:
@dcy if I understand your last comment correctly then it would apply to your specific application, but when adding something like this we need to think about the general case.

larshesel:
Thanks for the contribution, @dcy !

PR #509, opened 2017-09-20
Fix retained msg persistence/replication race issue

larshesel:
Persisting retained messages to plumtree is handled within the
`vmq_retain_src` gen_server by folding over the `?RETAIN_UPDATE` ets
table. After this fold is complete all objects in the `?RETAIN_UPDATE`
table are deleted.

This creates a race-condition as new retained messages are inserted
into the `?RETAIN_UPDATE` table *outside* of the `vmq_retain_src`
gen_server and it is therefore possible that rows are inserted between
the fold has finished and the objects are deleted, thus some updates
will be forgotten and never persisted (and therefore never replicated
to remote nodes!)

The naïve approach would be to move the insertions into the ETS table
into the `vmq_retain_src` gen_server, but having to call into a
gen_server for each retained message would likely quickly become a
bottleneck.

Another approach would be to only delete the entries from the
`?RETAIN_UPDATE` table we've persisted. But even that is prone to a
race-condition because a new update could happen between reading from
the table and deleting the entry. This race-condition would be much
more rare than the one above where all objects are deleted in bulk.

The solution is to use a counter. Each update  a counter
and every time an object is persisted the counter is
decremented. After the persist fold we delete all elements which have
a counter of zero (no pending updates).

This approach means we may write the same value more than once, which
is a good tradeoff compared to losing the value entirely.

### DIFF
--- a/apps/vmq_server/src/vmq_retain_srv.erl
+++ b/apps/vmq_server/src/vmq_retain_srv.erl
@@ -64,12 +64,12 @@ start_link() ->
 delete(MP, RoutingKey) ->
     Key = {MP, RoutingKey},
     ets:delete(?RETAIN_CACHE, Key),
-    ets:insert(?RETAIN_UPDATE, {Key}).
+    ets:update_counter(?RETAIN_UPDATE, Key, 1, {Key, 0}).
 
 insert(MP, RoutingKey, Message) ->
     Key = {MP, RoutingKey},
     ets:insert(?RETAIN_CACHE, {Key, Message}),
-    ets:insert(?RETAIN_UPDATE, {Key}).
+    ets:update_counter(?RETAIN_UPDATE, Key, 1, {Key, 0}).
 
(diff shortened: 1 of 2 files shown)


larshesel:
I just realized there is a corner case if the same key is written to many times. This would make the counter increase to some value X and only be decremented by one in each persist loop. That's a pretty nasty problem. Looking into how to overcome that.

larshesel:
Pushed a fix for the corner case mentioned (squashed)

edwinAtWiz:
@larshesel  since I am not familiar with erlang, as what you mentioned that, the approach means same value will be written more than once, will there be a chance for the second record is overriden by the first record?

larshesel:
@edwinAtWiz no. It's just we may end up writing the same value an extra time if we should trigger this race condition window. But this window has been reduced dramatically due to other changes as well.

PR #522, opened 2017-10-12
Convert the readme file into markdown

larshesel:
I got tired of looking at the reStructuredText as we use markdown everywhere else. Also changed the travis badge to only track the master branch (so it won't show a failed status when our other experimental/WIP branches are broken).


dergraf:
Lgtm

PR #546, opened 2017-11-27
Fix filtering sessions by topic issue

larshesel:
I'll try and add some tests tomorrow.


larshesel:
yeah - I wrote some more tests exposing more issues and I'm trying to fix those

larshesel:
Added some more tests and fixed some more bugs. Note this is built on top of #544

larshesel:
Fixed the stale-session test - it was relying on the output of a filter query which gave bad output and it broken when the filters were fixed...

PR #550, opened 2017-12-05
Added session force disconnect feature

dergraf:
Todo: 
- Add integration with vernemq_dev

### DIFF
--- a/apps/vmq_server/src/vmq_info_cli.erl
+++ b/apps/vmq_server/src/vmq_info_cli.erl
@@ -18,9 +18,11 @@
 
 register_cli() ->
     vmq_session_list_cmd(),
+    vmq_session_disconnect_cmd(),
 
     clique:register_usage(["vmq-admin", "session"], session_usage()),
-    clique:register_usage(["vmq-admin", "session", "show"], vmq_session_show_usage()).
+    clique:register_usage(["vmq-admin", "session", "show"], vmq_session_show_usage()),
+    clique:register_usage(["vmq-admin", "session", "disconnect"], vmq_session_disconnect_usage()).
 
 vmq_session_list_cmd() ->
     Cmd = ["vmq-admin", "session", "show"],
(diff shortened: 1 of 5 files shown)


dergraf:
@larshesel what do you think of this approach?

larshesel:
Looks pretty good to me - I noticed that at least one of the new tests failed on travis ...

apps/vmq_server/test/vmq_queue_SUITE.erl
```diff
@@ -236,6 +240,58 @@ queue_persistent_client_expiration_test(_) ->
     not_found = vmq_queue_sup_sup:get_queue_pid(SubscriberId),
     {ok, []} = vmq_lvldb_store:msg_store_find(SubscriberId).
 
+queue_force_disconnect_test(_) ->
+    Parent = self(),
+    SubscriberId = {"", <<"force-client-disconnect">>},
+    QueueOpts = maps:merge(vmq_queue:default_opts(), #{clean_session => false,
+                                                       max_offline_messages => 1000,
```
codeadict: Should this be a hardcoded 1000 or use the User's setting?

dergraf: you're right, for this test case it really doesn't matter.

apps/vmq_server/src/vmq_info_cli.erl
```diff
@@ -92,6 +128,7 @@ session_usage() ->
      "  Retrieve information on live sessions.\n\n",
```
larshesel: This should probably be changed to something else, like `Manage MQTT sessions`.

apps/vmq_server/src/vmq_info_cli.erl
```diff
@@ -92,6 +128,7 @@ session_usage() ->
...
      "  Sub-commands:\n",
      "    show        Show and filter running sessions\n",
+     "    disconnect  Forcefully disconnect a session\n",
```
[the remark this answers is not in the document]
drf: "cleanup  Forcefully clean up a session (will disconnect the client if connected)"

apps/vmq_server/src/vmq_queue.erl
```diff
@@ -520,6 +527,25 @@ handle_sync_event(get_sessions, _From, StateName, #state{sessions=Sessions} = St
     {reply, maps:keys(Sessions), StateName, State};
 handle_sync_event(get_opts, _From, StateName, #state{opts=Opts} = State) ->
     {reply, Opts, StateName, State};
+handle_sync_event({force_disconnect, DoCleanup}, _From, StateName,
+                  #state{id=SId, sessions=Sessions, offline=#queue{queue=OfflineQ}} = State) ->
+    %% Forces to disconnect all sessions and to cleanup all state
+    %% session has expired cleanup and go down
```
larshesel: The above line looks like an old cut'n'paste leftover. Should be removed.

apps/vmq_server/src/vmq_queue.erl
```diff
@@ -520,6 +527,25 @@ handle_sync_event(get_sessions, _From, StateName, #state{sessions=Sessions} = St
     {reply, maps:keys(Sessions), StateName, State};
 handle_sync_event(get_opts, _From, StateName, #state{opts=Opts} = State) ->
     {reply, Opts, StateName, State};
+handle_sync_event({force_disconnect, DoCleanup}, _From, StateName,
+                  #state{id=SId, sessions=Sessions, offline=#queue{queue=OfflineQ}} = State) ->
+    %% Forces to disconnect all sessions and to cleanup all state
```
larshesel: I think this would read nicer like this: `Forcefully disconnect all sessions and cleanup all state`

PR #561, opened 2017-12-20
Make it possible to keep MQTT username when using the proxy protocol

larshesel:
This feature was contributed by SoftAtHome (https://softathome.com/).


apps/vmq_server/src/vmq_listener_cli.erl
```diff
@@ -295,6 +299,13 @@ vmq_listener_start_usage() ->
      "WebSocket Options\n\n",
      "  --websocket\n",
      "      use the Websocket protocol as the underlying transport\n\n",
+     "PROXYv2 Options\n\n",
+     "  --proxy_protocol\n",
+     "      Enable PROXY v2 protocol for this listener\n",
+     "  --use_cn_as_username\n",
+     "      If PROXY v2 is enabled for this listener used this flag\n",
```
dergraf: typo "used"

dergraf: @larshesel ping

larshesel: Fixed.

PR #565, opened 2018-01-03
Remove unused time and random compatibility dependencies

larshesel:
As we're anyway not supporting OTP 17 we might as well get rid of the time compat libraries...


larshesel:
This should wait for erlio/plumtree#14 to be merged first then this PR should include the change to update plumtree.

PR #573, opened 2018-01-08
Strengthen check verifying if a remote node is available

larshesel:
This should make it more visible if the cluster configuration is
wrong or if there's a partial netsplit (only specific ports etc).

My initial thought here was that it's annoying that if for some reason the clusters cannot connect (`listener.vmq.clustering`) but the erlang distribution works, then `vmq-admin cluster show` would report true. That makes it hard for people to understand that something is wrong as it's only otherwise written into the logs.
 
Should this be for 1.3.0? It's not strictly a bugfix...

### DIFF
--- a/apps/vmq_server/src/vmq_cluster.erl
+++ b/apps/vmq_server/src/vmq_cluster.erl
@@ -167,7 +167,10 @@ check_ready([Node|Rest], Acc) ->
                        _ -> false
                    end,
     ok = vmq_cluster_node_sup:ensure_cluster_node(Node),
-    check_ready(Rest, [{Node, IsReady}|Acc]);
+    %% We should only say we're ready if we've established a
+    %% connection to the remote node.
+    IsReady1 = IsReady andalso vmq_cluster_node_sup:is_reachable(Node),
+    check_ready(Rest, [{Node, IsReady1}|Acc]);
 check_ready([], Acc) ->
     ClusterReady =
     case lists:keyfind(false, 2, Acc) of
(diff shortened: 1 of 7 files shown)


larshesel:
this breaks some tests which weren't ready for this stronger check, apparently ;)

apps/vmq_server/src/vmq_cluster.erl
```diff
@@ -167,7 +167,10 @@ check_ready([Node|Rest], Acc) ->
                        _ -> false
                    end,
     ok = vmq_cluster_node_sup:ensure_cluster_node(Node),
```
dergraf: we could possibly get rid of the rpc call alltogether, no?

dergraf:
1.2.3 is fine with me, it's a minor improvement/refactoring.

apps/vmq_server/src/vmq_cluster.erl
larshesel: The netsplit tests break if I remove the rpc call ... I'll have to figure out which assumption I'm breaking there.

larshesel: This was because removing the RPC call broke the netsplit tests as they only disrupt the erlang distribution (by changing the cookie) and don't do anything regarding the erlang cluster communications channel. For now I changed the netsplit code to do a `vmq_cluster_node_sup:del_cluster_node/1` and `vmq_cluster_node_sup:ensure_cluster_node/1` as well. I have a feeling this is not the right way as these calls might have side-effects which shouldn't happen when simulating a netsplit.

larshesel: thinking a bit more about it, removing the RPC calls defeats on purpose: namely that I'd like the check to check that both erlang distribution and the cluster comm are both working to declare the cluster ok. The question is, can we replace the rpc call with something less expensive?

Issue #554, opened 2017-12-09
Consider providing an in-memory datastore


picaoao:
We currently have a use case where we are considering using a mqtt broker to implement a presence system. In that scenario we don't need to actually persist in disk the retained messages, last-will,  as well as the topic subscriptions. Everything could work in-memory, and with that provide much higher performance, since we would be emitting all messages with retain flag set to true. We are aiming at very high nr of clients, configured with last-will, so this means that using leveldb and keep updating the retained message all the time and sync between clusters would be a killer in performance.

How hard would be to have everything running in a in-memory db ?
Emqtt implementation seems to have that option with retained messages by using mnesia in ram.

ioolkos:
Hi @picaoao thanks for your remarks, it's an interesting question.
There's 2 sides to this question:
- what is the best architecture to make the backend for persistence of any kind of state completely pluggable
- what is the fastest and simplest way for you to have everything as an in-memory DB as described

I have no easy answer to that. I think the architecture is not quite pluggable yet. And while I think the effort is low, it's probably not high-priority for us right now.

@dergraf @larshesel what do you think?

ioolkos:
A clarification: low effort is only the message store replacement. Any involvement with the plumtree protocol (as in the subscriber replication) is rather high effort.

PR #594, opened 2018-01-30
Make the message store configurable

larshesel:
So, this is a quick WIP I threw together to make it possible to decide which kind of message store one would like to use. Currently only two are supported, namely: `leveldb` and `none`.

I'm not completely sure about the `none` backend yet - it simply doesn't register any hooks for the backend, which means anything calling `vmq_plugin:only(msg_store_xxxx, ...` gets a `{error,no_matching_hook_found}` error back which effectively means no backend is present. We should probably look at all code paths touching the message store to verify they can handle that situation correctly. Another way would be to actually implement a `none` backend which would return proper answers and just ok any messages written into it. I'm not sure that's a good idea, as the code would then not be able to differentiate between persisted/not persisted...

Of course, the actual goal of this is to, somewhere down the road, be able to offer other message store backends than `leveldb` or `none`.

What it doesn't handle yet -  if is one wants to combine a layer on top of a backed. For example, some caching layer or a layer causing writes to be delayed. One could imagine that such a layer could be combined with any actual backend. I suppose that might be solved by being able to specify the layer as the backend and then configure the layer independently to use leveldb...

This PR would partially address #554

### DIFF
--- a/apps/vmq_server/priv/vmq_server.schema
+++ b/apps/vmq_server/priv/vmq_server.schema
@@ -1696,3 +1696,28 @@
          true;
     (Name) -> vmq_schema_util:file_is_readable(Name)
  end}.
+
+%% @doc Select the message store backend.
+{mapping, "message_store", "vmq_server.message_store", [
+                                                        {default, leveldb},
+                                                        {datatype, {enum, [leveldb, none]}},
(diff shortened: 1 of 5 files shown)


larshesel:
Closing this for now

Issue #595, opened 2018-01-30
Shared subscriber cannot receive messages via the other node after the cluster is recoverd


yuta-oxo:
### Environment

- VerneMQ Version: 1.3.0
https://bintray.com/artifact/download/erlio/vernemq/deb/xenial/vernemq_1.3.0-1_amd64.deb
- OS(server): Ubuntu 16.04 LTS on VirtualBox
- OS(client): macOS 10.13.3

### Expected behavior

Shared subscriber can receive messages via the other node after the cluster is recoverd.

1. Node A and node B are in a cluster.
2. Shared subscriber A (Sub_A) listens a topic on the node A.
3. Publisher A (Pub_A) sends messages to the topic on the node A.
4. Publisher B (Pub_B) sends messages to the topic on the node B.
5. Sub_A can receive the messages from Pub_A and Pub_B.
6. The node B is rebooted. The cluster is broken temporarily and then it is recoverd.
7. Pub_B reconnects to the node B.
8. Sub_A can receive the messages from Pub_A and Pub_B.

### Actual behaviour

Shared subscriber cannot receive messages via the other node after the cluster is recoverd.

1. Node A and node B are in a cluster.
2. Shared subscriber A (Sub_A) listens a topic on the node A.
3. Publisher A (Pub_A) sends messages to the topic on the node A.
4. Publisher B (Pub_B) sends messages to the topic on the node B.
5. Sub_A can receive the messages from Pub_A and Pub_B.
6. The node B is rebooted. The cluster is broken temporarily and then it is recoverd.
7. Pub_B reconnects to the node B.
8. Sub_A can receive the messages from Pub_A. But Sub_A cannot receive the messages from Pub_B.

Two nodes are in a cluster:
```
Node A: 192.168.56.64
Node B: 192.168.56.65
```

Node A's config:
```
allow_anonymous = on
allow_register_during_netsplit = on
allow_publish_during_netsplit = on
allow_subscribe_during_netsplit = on
allow_unsubscribe_during_netsplit = on
allow_multiple_sessions = on
max_inflight_messages = 20
max_online_messages = 1000
max_offline_messages = 1000
max_message_size = 0
upgrade_outgoing_qos = off
listener.max_connections = 10000
listener.nr_of_acceptors = 10
listener.tcp.default = 0.0.0.0:1883
listener.vmq.clustering = 192.168.56.64:44053
listener.http.default = 127.0.0.1:8888
listener.mountpoint = off
systree_enabled = on
[...]

PR #596, opened 2018-01-31
Fix trie initialization to include remote shared subscriptions

larshesel:
Fix for #595 

Will try and think of a test for this tomorrow to prevent this bug from reappearing.


larshesel:
Added a test.

PR #605, opened 2018-02-08
Generalize connect suite

larshesel:
I also started adding an `mqtts` group, but it didn't turn out as I wanted it. SSL is complex because it has so many different setups.

At least the code below fixes a couple of bugs and makes the connect suite faster by more than 50% (~13 of 23 secs).

To check that tests are truly independent, we could consider setting the `shuffle` property on the group.

Didn't add a changelog entry as I don't consider test-refactoring to be relevant to the users.


larshesel:
I just added a *very* basic websocket client and created a group in `vmq_connect_SUITE.erl` which runs all the tests using that client as well...

PR #620, opened 2018-02-28
Deliver messages to online shared subscribers first

### LINKED ISSUE #618: Persistence and shared subscriptions
- VerneMQ Version: 1.3.0
- OS: Server Debian - Client Windows

Status:
 - 2 Clients with shared subscriptions and persistent session subscribed to the same topic.
 - 1 Client publishing to the topic.

At first you see how the messages are received randomly between both clients. Everything is fine. When you close one client, messages are still sent to closed clients like they were active, so only half of the messages are processed actually.

As per [this message](https://github.com/erlio/vernemq/issues/230#issuecomment-274321440), closed clients should not receive messages and only be sent to active clients.

When you disable persistence, everything works as expected.

Small test that reproduces the problem with a default VerneMQ installation.
```
let mqtt = require('mqtt');

messages = 10;
test_topic = 'strangeTopic';
let client1 = {
    url: 'mqtt://10.2.3.31:1883',
    options: { clean: false, clientId: 'test_client1' },
}
let client2 = {
    url: 'mqtt://10.2.3.31:1883',
    options: { clean: false, clientId: 'test_client2' },
}

let publisher = {
    url: 'mqtt://10.2.3.31:1883',
    options: {}
}

let conn1 = mqtt.connect(client1.url, client1.options);
let conn2 = null;

conn1.on('connect', (err, res) => {
    conn1.subscribe(`$share/test/${test_topic}`, (err, res) => {
        conn2 = mqtt.connect(client2.url, client2.options);
        conn2.on('connect', (err, res) => {
            conn2.subscribe(`$share/test/${test_topic}`, (err, res) => {
                conn1.end();
                publish();
            });
            conn2.on('message', (topic, payload, packet) => {
                console.log(`message received on client2 with topic: ${topic} and payload: ${payload}`)
            })
        });
    });
    conn1.on('message', (topic, payload, packet) => {
        console.log(`message received on client1 with topic: ${topic} and payload: ${payload}`)
    });
});

let publish = () => {
    let conn3 = mqtt.connect(publisher.url);
    conn3.on('connect', (err, res) => {
        // a message per second
        for (let i = 0; i < messages; i++) {
            setTimeout(() => {
                conn3.publish(test_topic, 'test payload', (err, res) => {
                    console.log(`published ${i + 1} messages`);
                 });
            }, (i + 1) * 1000);
        }
    })
}
```

larshesel:
Fix bug preventing the delivery of messages to a shared subscriber group from being delivered to any online members first before attempting to deliver the messages to offline members.

Fixes #618

### DIFF
--- a/apps/vmq_server/src/vmq_shared_subscriptions.erl
+++ b/apps/vmq_server/src/vmq_shared_subscriptions.erl
@@ -57,7 +57,7 @@ publish_online(Msg, Subscribers) ->
                       _ ->
                           Acc
                   end
-          end, Subscribers, [])
+          end, [], Subscribers)
     catch
         done -> ok
     end.
--- a/apps/vmq_server/test/vmq_publish_SUITE.erl
+++ b/apps/vmq_server/test/vmq_publish_SUITE.erl
@@ -56,7 +56,8 @@ groups() ->
          not_allowed_publish_close_qos0_mqtt_3_1_1,
          not_allowed_publish_close_qos1_mqtt_3_1_1,
          not_allowed_publish_close_qos2_mqtt_3_1_1,
-         message_size_exceeded_close
+         message_size_exceeded_close,
+         shared_subscription_online_first
         ],
     [
      {mqtt, [shuffle,sequence], Tests}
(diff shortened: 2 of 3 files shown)


kellycampbell:
I'm still seeing this issue with the latest nightly (2018-03-08).

I'm attaching my test scripts. I run 1 publisher script and 4 subscriber scripts and then kill/restart the sub scripts as desired. What I saw was when I had only one subscriber running, and it was not receiving all messages, then I restarted one of the other subscribers and it immediately got the missing msgs redelivered.

(It won't let me attach .sh files so copy/pasting them here)

publish.sh
```bash
#!/bin/bash

SERVER=mqttserver
PORT=1883
TOPIC=test_topic
COUNTER=0

while [ true ] ; do
    ((COUNTER++))
    TS=$(date "+%Y-%m-%d %H:%M:%S")

    mosquitto_pub -h $SERVER -p $PORT -q 2 -t "${TOPIC}/${COUNTER}/${TS}" -m ""
    if [ $? -ne 0 ] ; then
        echo "publish failed! ${TOPIC}/${COUNTER}/${TS}"
    else
        echo "${TOPIC}/${COUNTER}/${TS}"
    fi
    sleep 3
done
```

subscribe.sh
```bash
#!/bin/bash

# Usage: subscriber.sh <clientid>

SERVER=mqttserver
PORT=1883
SHARENAME=test_share
TOPIC=test_topic

echo "Subscribing to \$share/${SHARENAME}/${TOPIC}/#"
mosquitto_sub -c -h $SERVER -p $PORT -v -q 2 -i $1 -t "\$share/${SHARENAME}/${TOPIC}/#" | ts '[%Y-%m-%d %H:%M:%S]'
if [ $? -ne 0 ] ; then
    echo "subscribe failed! \$share/${SHARENAME}/${TOPIC}/#"
fi
```

larshesel:
Hi - to see the node a client is connected to you can pass the `--node` flag to `vmq-admin session show`. See also `vmq-admin session show --help`.

It looks like the latest docker build contains VerneMQ 1.3.0 and not the nightly... not sure why - do you have the possibility to build directly from master and test that?

kellycampbell:
Thanks. I failed to check the --help on the session show.

We're using a custom docker build. See https://github.com/kellycampbell/docker-vernemq/tree/k8s-fixes

The above testing used this .deb package:

```
# 1.3.0-f96a66aa = Nightly version from 20180308
ENV VERNEMQ_VERSION 1.3.0-f96a66aa

ADD https://bintray.com/artifact/download/erlio/vernemq/deb/jessie/vernemq_$VERNEMQ_VERSION-1_amd64.deb /tmp/vernemq.deb
```

larshesel:
In the test are all clients connect to the same node, or how are they distributed among the nodes?

kellycampbell:
It was random via an AWS classic ELB. Since each publish is a new connection, the publishing node is  random for each message.

larshesel:
(sorry, posted this in the wrong thread a few moments ago, here it is again)

I believe what's causing this is that the publisher is using the default policy of prefer_local. If there is an offline (or online) session on the node where the publisher is connected, then all messages will be delivered to that session and clients that are connected to the other node will not receive any messages.

jeevansirela:
Hi @larshesel, 
I am also facing a similar issue, cs = false(persistence), and I have enabled the shared subscription policy to random too. But my publisher is still saying prefer_local in logs. The strange thing is it's not happening to another shared topic. But happening to this "will" topic.

PR #633, opened 2018-03-16
Plugin refactoring part 1

larshesel:
This is a first PR of two which generalizes the plugin system to allow for injecting compatibility code around a hook.

This PR just changes the internal API to allow for a general option argument instead of just passing in paths.


larshesel:
Maybe I should've updated the readme instead of removing it... Though I don't think anyone ever sees that documentation as it's a bit too well hidden..

PR #636, opened 2018-03-19
Fix rpi32 build

larshesel:
Needs a  changelog item.


larshesel:
IIRC it's because the directories weren't generated in the compile step (it used to work with earlier rebar3 versions), but when we get to the release step we know the directories are there and can link them. Or something like that...

PR #637, opened 2018-03-19
Plugin refactoring part 2

larshesel:
This builds on top of #633 which should be reviewed and merged first (this PR then has fewer changes)

This PR makes it possible to specify compatibility data for an application plugin. This compatibility data (basically a module and a function) can be used to wrap existing plugin hooks, so they can be converted to the format of a new hook by wrangling parameters and return values.

The PR also refactors the internal representation of plugin hooks to use records instead of matching on different tuple sizes.

We will need this to be able to reuse old plugins in an MQTTv5 context.

Note: before this PR can be merged I have to do some experiments actually using this to make sure the API works... But a basic review would be nice.

### DIFF
--- a/apps/vmq_plugin/src/vmq_plugin.hrl
+++ b/apps/vmq_plugin/src/vmq_plugin.hrl
@@ -1,0 +1,12 @@
+-record(hook, {
+          name        :: atom(),
+          module      :: module(),
+          function    :: atom(),
+          arity       :: non_neg_integer(),
+          compat      :: {Name :: atom(),
+                          Mod   :: module(),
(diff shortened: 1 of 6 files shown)


larshesel:
rebased on top of master

larshesel:
Pushed a commit fixing `vmq-admin plugin show` which was broken. There should be some more infomation added for plugins with compatibility hooks for `vmq-admin plugin show` but that is still not there.

larshesel:
Closing this as we'll just bring this in when merging the mqttv5-preview branch.

PR #638, opened 2018-03-19
Initial version of a simple VerneMQ clustering status page

dergraf:
screenshot from 2018-03-19 20-22-12
The status page shows the following information:
- Node information
  - Peercount, how many cluster members this node can communicate with
  - Connect Rate (sampled and averaged at 10 second intervals)
  - Publish In/Out Rate (sampled and averaged at 10 second intervals)
  - Msg drop Rate (sampled and averaged at 10 second intervals)
  - Nr. of clients (online/offline)
  - Nr. of queued messages
  - Nr. of subscriptions
  - Nr. of retained messages
  - Running Protocols (a selection of MQTT, MQTTS, MQTTWS)
  - VerneMQ Release version

- Cluster information
  - Cluster size
  - Aggregated Node information (see above)

- Cluster issues, shows issues, that might degrade overall stability
  - Warns if cluster communication uses 0.0.0.0 interface
  - Warns if VerneMQ release version differs among cluster nodes
  - Warns if a cluster node isn't reachable by any other node
  - Informs if a MQTT(S|WS) listener isn't running

Technology:
- Bootstrap, Mustache.js, JQuery


larshesel:
I think this looks great!

Do we have to declare something (beta/not api) to be able to change this in the future?  Or is it the general expectation that we can change this however we like in a minor version?

One enhancement idea would be to explain why it is a warning to use the 0.0.0.0 interface and why that is a potential problem. Maybe we could add links to the wiki or docs for explaining this?

Should it be possible to disable this feature?

A changelog item should be added.

codeadict:
Is this merged into the new release?

larshesel:
No, as today's release was a bugfix it was not included. It'll be part of 1.4.0.

larshesel:
I'll should be in the nightly build though (which is just 1.3.1 + this PR).

Issue #619, opened 2018-02-27
Crashes during API call to /api/v1/session/show


alisoviy-mev:
### Environment

- VerneMQ Version: 1.3.0
- OS: Ubuntu 16.04.4 LTS

We have a VerneMQ installed on AWS instance alongside with our application.
Every 10 seconds our app make API call to broker to get a list of sessions:
`http://localhost:8001/api/v1/session/show?--limit=200&--rowtimeout=10000`
Broker have up to 30 connection over TCP and up to 10 connections over WS.

As configuration we added this section to end of vernemq.conf:
```
########## Start ##########

listener.http.default=127.0.0.1:8001
listener.tcp.default=0.0.0.0:1883
listener.ws.default=0.0.0.0:1884

plugins.vmq_webhooks=on
vmq_webhooks.mywebhook1.hook=auth_on_register
vmq_webhooks.mywebhook1.endpoint=http://127.0.0.1:8080/authonregister
vmq_webhooks.mywebhook2.hook=auth_on_publish
vmq_webhooks.mywebhook2.endpoint=http://127.0.0.1:8080/authonpublish
vmq_webhooks.mywebhook3.hook=on_client_wakeup
vmq_webhooks.mywebhook3.endpoint=http://127.0.0.1:8080/onclientwakeup
vmq_webhooks.mywebhook4.hook=on_client_gone
vmq_webhooks.mywebhook4.endpoint=http://127.0.0.1:8080/onclientgone
vmq_webhooks.mywebhook5.hook=on_client_offline
vmq_webhooks.mywebhook5.endpoint=http://127.0.0.1:8080/onclientoffline
vmq_webhooks.mywebhook6.hook=auth_on_subscribe
vmq_webhooks.mywebhook6.endpoint=http://127.0.0.1:8080/authonsubscribe

graphite_enabled=on
graphite_host=127.0.0.1
graphite_port=2003
graphite_interval=10000
graphite_prefix=vernemq

########### End ###########
```

### Expected behavior

- Sessions list to be returned every time.

### Actual behaviour

Unfortunately I cannot provide any steps for reproduce, because we didn't found any specific actions before errors. And we hope that someone can provide us any info about how we can get additional info about the errors.

- Sometimes we had crash reports during API call (to get sessions list). It happens up to 15 times per day, and we cannot reproduce this errors.

crash.log:
```
2018-02-26 21:05:52 =ERROR REPORT====
** Generic server <0.21280.2> terminating 
[...]

PR #649, opened 2018-04-04
Handle `normal` exit when fetching from `vmq_ql_query`

dergraf:
might be related to what has been reported here https://github.com/erlio/vernemq/issues/619


larshesel:
Looks good - with this patch I can no longer reproduce the crash (exit normal) in #619.

Can you rebase it on top of latest from master and add a changelog item? Then I'd merge it.

PR #674, opened 2018-04-30
Be backwards compatible with OTP18

larshesel:


larshesel:
Notice the tests fail because some of them are implemented to fail (`throw(not_implemented)`). The next PR ( #675 )changes that so those tests will be skipped instead

larshesel:
closed via #675

PR #679, opened 2018-05-02
Add MQTTv5 receive maximum feature

dergraf:


apps/vmq_server/src/vmq_mqtt5_fsm.erl
```diff
@@ -418,8 +440,7 @@ connected(#mqtt5_subscribe{message_id=MessageId, topics=Topics, properties=_Prop
             _ = vmq_metrics:incr_mqtt_suback_sent(),
             {State, [Frame]};
         {error, not_allowed} ->
-            %% allow the parser to add the 0x80 Failure return code
-            QoSs = [not_allowed || _ <- Topics],
+            QoSs = [16#87 || _ <- Topics], % not authorized
```
larshesel: Use `?M5_NOT_AUTHORIZED` instead of `16#87`.

apps/vmq_server/src/vmq_mqtt5_fsm.erl
```diff
@@ -295,12 +309,8 @@ connected(#mqtt5_publish{message_id=MessageId, topic=Topic,
...
     end,
     case Ret of
-        {error, not_allowed} ->
-            %% TODOv5 this will tell the client that the publish
-            %% wasn't allowed. Is it a good idea to give the client
-            %% this information or is it better to just drop the
-            %% connection?
-            terminate(not_authorized, State);
+        {error, recv_max_exceeded} ->
+            terminate(?RECEIVE_MAX_EXCEEDED, State);
```
larshesel: Can't we still get `{error, not_allowed}`  here, for instance from a plugin which doesn't allow the publish?

dergraf: I think I've covered all occurrences.

PR #683, opened 2018-05-04
Prepare message store to be able to use versioned message store values.

dergraf:


larshesel:
I just had a thought and was wondering, if instead of tagging each version, we could have a single k/v pair which says what the version is for the entire k/v store. That would save some space, but would likely be a bit less robust.

PR #684, opened 2018-05-04
Implement retain as publish subscribe flag

larshesel:
This is the first part of implementing the subscription options and this implements the Retain as Publish flag.

Currently working on implementing the 'no local' and 'retain handling' flags, but I though it would be nice to review this first as it's already quite large as is and touches a lot of different places in the code base.

### DIFF
--- a/apps/vmq_server/src/vmq_reg.erl
+++ b/apps/vmq_server/src/vmq_reg.erl
@@ -66,8 +66,8 @@
 
 
 -spec subscribe(flag(), subscriber_id(),
-                [{topic(), qos()}]) -> {ok, [qos() | not_allowed]} |
-                                       {error, not_allowed | not_ready}.
+                [subscription()]) -> {ok, [qos() | not_allowed]} |
+                                     {error, not_allowed | not_ready}.
 subscribe(false, SubscriberId, Topics) ->
     %% trade availability for consistency
     vmq_cluster:if_ready(fun subscribe_op/2, [SubscriberId, Topics]);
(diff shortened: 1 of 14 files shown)


apps/vmq_server/src/vmq_reg.erl
```diff
@@ -597,12 +619,21 @@ fold_subscribers(FoldFun, Acc) ->
               FoldFun(SubscriberId, Subs, AccAcc)
       end, Acc).
 
--spec add_subscriber([{topic(), qos() | not_allowed}], subscriber_id()) -> ok.
+-spec add_subscriber([{topic(), qos() | not_allowed} |
+                      {topic(), qos() | not_allowed, map()}], subscriber_id()) -> ok.
```
dergraf: Shouldn't the spec looks like: 
```erlang
-spec add_subscriber([{topic(), qos() | not_allowed} |
                      {topic(), {qos() | not_allowed, map()}}], subscriber_id()) -> ok.
```

apps/vmq_server/src/vmq_subscriber.erl
```diff
@@ -173,13 +173,19 @@ get_node_subs(Node, Subs) ->
 -spec fold(function(), any(), subs() | changes()) -> any().
 fold(Fun, Acc, Subs) ->
     lists:foldl(fun({Node, _, NSubs}, AccAcc) ->
-                        lists:foldl(fun({Topic, QoS}, AccAccAcc) ->
-                                            Fun({Topic, QoS, Node}, AccAccAcc)
-                                    end, AccAcc, NSubs);
+                        lists:foldl(
+                          fun({Topic, QoS}, AccAccAcc) ->
+                                  Fun({Topic, QoS, Node}, AccAccAcc);
+                             ({Topic, QoS, Opts}, AccAccAcc) ->
```
dergraf: Rename QoS to SubInfo, and remove never matching case on linen 179

apps/vmq_server/src/vmq_subscriber.erl
```diff
@@ -173,13 +173,19 @@ get_node_subs(Node, Subs) ->
...
+                                  Fun({Topic, {QoS, Opts}, Node}, AccAccAcc)
+                          end, AccAcc, NSubs);
                    ({Node, NChanges}, AccAcc) ->
-                        lists:foldl(fun({Topic, QoS}, AccAccAcc) ->
-                                            Fun({Topic, QoS, Node}, AccAccAcc)
-                                    end, AccAcc, NChanges)
+                        lists:foldl(
+                          fun({Topic, QoS}, AccAccAcc) ->
+                                  Fun({Topic, QoS, Node}, AccAccAcc);
+                             ({Topic, QoS, Opts}, AccAccAcc) ->
```
dergraf: See Comment above

apps/vmq_server/src/vmq_reg.erl
larshesel: Fixed

apps/vmq_server/src/vmq_subscriber.erl
larshesel: Fixed

apps/vmq_server/src/vmq_subscriber.erl
larshesel: Fixed

PR #688, opened 2018-05-08
Plumtree abstraction

dergraf:
This PR abstracts the use of Plumtree as a cluster membership service as well as a metadata storage layer.

How is it done:
New `vmq_plugin` hooks are defined that generalize the interaction with a cluster membership service as well as metadata storage service.  A new application `vmq_plumtree` is added that implements all those hooks. All calls to the `plumtree` app are refactored to take the round trip over the `vmq_plugin:only` call.


apps/vmq_plugin/src/vmq_plugin_mgr.erl
```diff
@@ -390,7 +398,7 @@ check_updated_plugins(Plugins, State) ->
             ok = init_plugins_cli(CheckedPlugins),
             ok = start_plugins(CheckedPlugins),
             ok = compile_hooks(CheckedPlugins),
-            {ok, handle_deferred_calls(State#state{ready=true, plugins=CheckedPlugins})};
```
dergraf: This was an actual bug in `vmq_plugin_mgr` which actually never got triggered as we have deferred all calls.

apps/vmq_server/src/vmq_cluster_mon.erl
```diff
@@ -127,7 +127,7 @@ handle_info({nodeup, Node}, State) ->
     vmq_cluster:recheck(),
     {noreply, State};
 handle_info({gen_event_EXIT, vmq_cluster, _}, State) ->
-    plumtree_peer_service_events:add_sup_handler(vmq_cluster, [self()]),
```
dergraf: to pass `self()` as an arg to `add_sup_handler` was an actual bug which would result in a bad match error.

larshesel: This would be because the `vmq_cluster:init/1` only accepts the empty list `[]` as an argument, right? (not that versed in `gen_event` ... )

PR #692, opened 2018-05-09
Pass the path as proplist when starting a user plugin

rbino:
This is needed after bcad8893bf5ebdfa08eb16cfaaeaa72eae148bf9


larshesel:
Could I get you to add a line to the changelog.md file mentioning the fix?

rbino:
Done

rbino:
I'm not sure why the CI fails

larshesel:
We have some tests which are less deterministic than we'd want so sometimes they fail - something to look into at some point. Anyway, all tests passed so I'll merge it to master.

Thanks a lot for this fix!

PR #707, opened 2018-05-17
Revert "Upgrade rebar3 (#695)"

larshesel:
This reverts commit 9a5c4c39009cd967b61ab62627322e413987012d.

The rebar3 upgrade had a strange side-effect (cause still unknown) which made it impossible to load plugins. Until we figure out what that issue was let's revert back to the old rebar3 version (3.5.3 -> 3.3.6).


rbino:
Tested and this fixes the issue for me

Issue #661, opened 2018-04-18
sudo vernemq top


ioolkos:
I guess we could add etop, so that folks could use the quick `sudo vernemq top` command. Was there any specific reason we left this out?

larshesel:
I know of no reason why we left that out. We could include it in 1.4.0.

codeadict:
I will add this, promise :)

codeadict:
Working on this, I think it will make more sense to go with `vmq-admin top`. What are your thoughts @ioolkos ?

ioolkos:
I'd like to actually support what the `vernemq` script advertises. Which doesn't exclude adding top to `vmq-admin` too, of course.. and possibly extende the filtering or sorting opts there.

```
➜  vernemq git:(master) ✗ vernemq

Usage: vernemq «command»
where «command» is one of the following:
    { help | start | stop | restart | ping | console | attach
      attach-direct | ertspath | chkconfig | escript | version | getpid
      top [-interval N] [-sort { reductions | memory | msg_q }] [-lines N] } |
      config { generate | effective | describe VARIABLE } [-l debug]

Run `vernemq help` for more detailed information.

```

codeadict:
That makes lot of sense

larshesel:
As it's a debug tool I think it's fine to only have it in the `vernemq` command. It's likely not a tool one would use often(hopefully never), but nice to have in a pinch.

codeadict:
Pull requests submitted! 🎉

larshesel:
I think I'd keep it in the erlio/node_package repo for now. Then it's easy to understand where the runner script comes from and if upstream changes at some point we don't have to remember to bring those things over to our copy of the runner script. So I'd say we should keep it in the erlio/node_package repo.

PR #713, opened 2018-05-21
Load the observer that contains etop with the release

codeadict:
This is the fix for #661, accompanying Pull request was submitted the `node_package` repository at https://github.com/erlio/node_package/pull/1. Already tested and working great locally.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -62,6 +62,8 @@
   graphite api key from being set in the `vernemq.conf` file.
 - Bugfix: WebHooks Plugin. Close the Hackney CRef so that the socket is given
   back to the Hackney pool, for the case of non-200 HTTP OK status codes.
+- Bugfix: Fix `vernemq top` command(like Erlang's etop) for presenting information
+  about node processes (#661).
 
 ## VerneMQ 1.3.0
 
--- a/rebar.config
+++ b/rebar.config
@@ -87,6 +87,7 @@
     {syslog, load},
     {lager_syslog, load},
     {runtime_tools, load},
+    {observer, load},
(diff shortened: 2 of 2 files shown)


ioolkos:
@codeadict thanks!
I was thinking that `etop` could be added independantly, I'll check if that's possible. I guess the full Observer (as a GUI) is not really needed, at least it wasn't included in the packages so far.

dergraf:
etop should be provided as a standalone erl file. we can't include observer due to the WX dependency.

codeadict:
Ah very True, that is why Basho folks used to rip `etop` from OTP and put it in the patches(this case erlio-patches) directory because there is no way to include etop independently.

dergraf:
exactly

dergraf:
remove `vernemq top` ;)

dergraf:
@codeadict would you be fine with removing top altogether?

codeadict:
Sure, I will remove it

codeadict:
I think recon is good enough to discover top processes

codeadict:
And as i mentioned https://github.com/zhongwencool/observer_cli uses recon and is something i can include really easy with no WX dependencies

larshesel:
I'm not sure if we should include that. If I want to insect a node I usually fire up another node where observer is installed/available and connect via observer to the node I want to investigate. Something like observer_cli could be installed as a plugin if needed, I guess. I'm just thinking along the lines that we shouldn't include anything unless we have a good reason to.

codeadict:
Closing this now, the real fix will be made on erlio/node_package#1.

PR #716, opened 2018-05-23
Fix issue with disconnecting a client with a specific mountpoint

larshesel:


ioolkos:
I think this solution is 'good enough'. Nobody remembers the default mountpoint as "" and then goes --mountpoint="" anyway. at least I hope so.

PR #724, opened 2018-05-30
Rework hooks + demo plugin + demo plugin test suite

larshesel:
The initial goal was to create a demo-plugin, but instead I reworked the plugins hooks to take properties into account and changed the fsm, tests etc to make that work.

There *is* a demo plugin `vmq_mqtt5_demo_plugin` which implements the v1 hooks and just logs the parameters. The next step would be to build out the demo plugin together with a test-suite which would test different aspects of MQTTv5 like reason codes, reason strings, user properties etc.


codeadict:
Should this go into VerneMQ itself or as a separate repo?

larshesel:
This goes into the mqtt5-preview branch which will later be merged into the master branch. We just need to be fairly sure that 1) we don't break anything 2) we don't do backwards incompatible changes.

larshesel:
Hey @dergraf, I think this is done and can be reviewed.

@ioolkos I've been thinking a bit about the hook names and after our last discussions I'm maybe starting to lean towards giving the new mqtt5 hooks a name which reflects that they are mqtt5 fsm hooks. My reasoning is that if we later start introducing new hooks (_v2) in the mqtt3/4 fsm, then it'll get hard to understand the difference between for instance `auth_on_register` (mqtt 3/4), `auth_on_register_v1` (mqtt 5) and `auth_on_register_v2` (mqtt 3/4).

I'm lacking good names for the new mqtt5 hooks: `auth_on_register_mqtt5`, `auth_on_register_m5`, `mqtt5_auth_on_register`, `m5_auth_on_register`? Something which can then later be postfixed with a version should we want to change that particular hook to a new format. So maybe `mqtt5_auth_on_register` isn't too bad: `mqtt5_auth_on_register_v1`.

Anyway, this is somethinig we can rather easily change and I think it should be done after this PR if we decide to do it...

Suggestions very welcome.

apps/vmq_mqtt5_demo_plugin/src/vmq_mqtt5_demo_plugin.erl
```diff
@@ -0,0 +1,138 @@
...
+                      [{[<<"suback">>,<<"withprops">>], _}] = Topics,
+                      _Properties) ->
+    {ok, #{topics => Topics,
+           properties => #{?P_USER_PROPERTY => [{<<"key">>, <<"val">>}],
+                           ?P_REASON_STRING => <<"successful subscribe">>}}};
+auth_on_subscribe_v1_(_Username,_SubscriberId,_Topics,_Properties) ->
+    ok.
+
+on_subscribe_v1(Username,SubscriberId,Topics)->
+    ?LOG([auth_on_subscribe_v1,Username,SubscriberId,Topics]),
```
dergraf: should be `on_subscribe_v1`

Issue #726, opened 2018-06-01
vernemq does not retry bridging connection after timeout 


vnejindal:
### Environment

- VerneMQ Version: 1.2.2 
- OS: Ubuntu 16.04
- Erlang/OTP version (if building from source):

### Expected behavior
The vernemq should keep retrying in case the bridging connection breaks or timesout. 

### Actual behaviour

Hi All, 
I have one query about an issue with following vmq MQTT bridging error trace. This trace was generated and vmq bridging stopped working. vmq had to be restarted to make it work again. Version: 1.2.2 on Ubuntu 16.04.

 Please let me know the reason for this? Thanks in advance.
```
2018-05-31 07:27:32.742 [error] <0.29708.0> gen_fsm <0.29708.0> in state connected terminated with reason: no function clause matching vmq_bridge:handle_info({tcp_error,#Port<0.38563>,etimedout}, {coord,<0.29639.0>}) line 167
2018-05-31 07:27:32.742 [error] <0.29708.0> CRASH REPORT Process <0.29708.0> with 2 neighbours exited with reason: no function clause matching vmq_bridge:handle_info({tcp_error,#Port<0.38563>,etimedout}, {coord,<0.29639.0>}) line 167 in gen_fsm:terminate/7 line 559
2018-05-31 07:27:32.742 [error] <0.22803.0> Supervisor vmq_bridge_sup had child {vmq_bridge,"#########",1883} started with vmq_bridge:start_link("#######", 1883, {vmq_reg,direct_plugin_exports,[vmq_bridge]}) at <0.29639.0> exit with reason no function clause matching vmq_bridge:handle_info({tcp_error,#Port<0.38563>,etimedout}, {coord,<0.29639.0>}) line 167 in context child_terminated (edited)
```

larshesel:
Hi, thanks for opening the issue. We'll try to take a look as soon as possible.

larshesel:
Hi @vnejindal , I created a PR for this in #728 - it still needs to pass review, but if you have the possibility to test it that would be awesome.

larshesel:
The PR was merged to master and is available in the nightly builds

vnejindal:
Thanks @larshesel . Waiting it to be part of upcoming release...

PR #728, opened 2018-06-05
Fix vmq_bridge reconnect issue

larshesel:
This should fix the issue in #726. Should we add some extra logging for these cases?


ioolkos:
This is a general fix for the mqtt client, right? vmq_mzbench uses the old repo as a dependency. I wonder how I could switch to the new version. vmq_mzbench uses https://github.com/erlio/vmq_commons/blob/master/src/gen_emqtt.erl currently.

larshesel:
vmq_commons is deprecated and I don't think there is a way to directly use vmq_commons from the vernemq repo. I guess we need to figure out how we wanna support things like vmq_mzbench in the future. One solution is that we put the client into its own repo again.

larshesel:
I guess one way to do that would be to put the client module into the cap'n nemo repo and declare it unstable. Then vmq_mzbench could depend on that but we don't have to worry about backwards compat towards other users leaving us free to completely redesign the client as cap'n nemo is/was supposed to become a much better client than the old one ever was.

PR #730, opened 2018-06-08
RFC - Generalize metrics

larshesel:
The purpose of this PR is to generalize the metrics system in a way
   such that we can:

   1. Have metrics with the same conceptual name, but still be able to
      distinguish them via tags describing for example a
      context. Example: `mqtt_publish_received` for MQTT 5 and MQTT
      3.1.1 - one should be able to ask for any mqtt publish messages
      received or just for the ones relevant for a specific protocol
      version.

   2. Metrics with the same conceptual name but different tags are
      aggregated (summarized) if no specific tag was requested.

   3. It is possible to retrieve all metrics without aggregations
      applied. In this case tag information will also be delivered to
      the receiver. Unaggregated metrics is required for monitoring
      systems like Prometheus where it should be possible to do the
      aggregation manually or only inspect events belonging to a
      specific tag if desired.

   4. Some metrics cannot be aggregated via summation due to their
      nature. One example are the scheduler utilization gauges as
      these are averages over time. Of course these are unique and
      have different conceptual names
      (`system_utilization_scheduler_1`,
      `system_utilization_scheduler_2`) and as such will not be
      summarized. In general metrics which should never be aggregated
      need to have unique conceptual names!

   5. Other kinds of aggregation than summation could be
      constructed. For instance instead of summarizing some kind of
      general function could be calculated over the metrics in
      question. For example an average could be used: instead of
      getting the total number of `mqtt_publish_received` events for
      all tags, one could calculate the average instead. It's not
      clear what the benefit would be here and as no obvious examples
      were thought of only aggregation via summation is used. It
      should also be mentioned that these other aggregation mechanism
      can also be applied later, for instance using a monitoring
      system like Prometheus.

Todos:

- [x] add tag support to the cli
- [x] provide tag information to the prometheus exporter
- [x] provide tag information to the grafana exporter
- [x] provide tag information to the $SYSTREE exporter
- [x] create a metrics manager process where we can register used tags and register new metrics (and it can be checked if it already exists)

### DIFF
--- a/apps/vmq_server/src/vmq_metrics_http.erl
+++ b/apps/vmq_server/src/vmq_metrics_http.erl
@@ -15,6 +15,8 @@
 -module(vmq_metrics_http).
 -behaviour(vmq_http_config).
 
+-include("vmq_metrics.hrl").
+
 -export([routes/0]).
 -export([init/3,
          handle/2,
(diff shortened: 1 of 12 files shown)


larshesel:
@dergraf, I think this is starting to get ready. Would you mind taking a look? Note the original comment in the PR which was updated.

apps/vmq_server/src/vmq_metrics_http.erl
```diff
@@ -36,33 +38,42 @@ terminate(_Reason, _Req, _State) ->
...
+    [line(BinMetric, Node, Tag, BinVal) || Tag <- Tags].
 
+line(BinMetric, Node, BinVal) ->
+    [BinMetric, <<"{">>, label(<<"node">>, Node), <<"} ">>, BinVal, <<"\n">>].
 
+line(BinMetric, Node, Tag, BinVal) ->
+    [BinMetric,
+     <<"{">>,
+     label(<<"node">>, Node), <<",">>,
+     label(<<"tag">>, atom_to_binary(Tag, utf8)),
```
dergraf: Maybe this is overkill, but in Prometheus you usually have multiple labels per metric while this approach would limit the labels to `node=xy` and `tag=xy`. I think for our case it would be definitive enough but it would look nicer if the name `tag` isn't fixed but rather dependent on the metric... let's say for the `mqtt_*` metrics it could be `mqtt_version=[4|5]` it could even contain more labels e.g. for specifying transport protocol `[ws|wss|tcp|tls]`... but as said it's probably overkill.

larshesel: If we start differentiating the labels then we'd have to come up with sensible differentiators (which we'd need to store together with the metrics definitions).

So, for example the metric `counter.mqtt_publish_received` would have the tags `mqtt_version=4`. This could be stored as a tag k/v pair `{mqtt_version,4}` where the metric is defined and would play well with prometheus. How would this map into the graphite exporter? Maybe sth like `mqtt.publish.received.mqtt_version_4` where the k/v pair is concatenated with `_` and a metric line is created for each such k/v pair?

Also, I'm not sure how would we specify this on the CLI: `vmq-admin metrics show --tags=k=v` doesn't work as it's not currently allowed to have more than one `=` in a flag. I guess `vmq-admin metrics show --mqtt_version=4` would be the natural choice. This would work a bit like the `vmq-admin sesion show` flags.

Then the question is how this information should be displayed on the cli if not aggregating. This could be similar to the output generated for graphite, I guess (aggregation is default, so passing no flags wouldn't display any tags, only the metrics names.

What do you think?

dergraf: I think for the proposed solution for Graphite would work.. but we could also decide to do the aggregation by default for Graphite... 

I like the idea of having `vmq-admin metrics show --mqtt_version=4 --other=this ..` I think we should default to aggregate if multiple metrics for the same name with different labels exist. The case where one explicitly wants no aggregation with something like `vmq-admin metrics show --disable-aggregation` we could print them similarly as Prometheus would do e.g. inside the 'name' column `metric_name{label1=val1, label2=val2..}` and inside the `value` column the value of the metric.
 
What do you think?

larshesel:
just rebased on top of master

dergraf:
We need to remove Label support for the Graphite exporter, or make it configurable. Otherwise we loose backward compatibility.

larshesel:
@dergraf I just added a commit making labels in graphite configurable.

larshesel:
alright, I'll squash and rebase this and merge it!

PR #731, opened 2018-06-12
Improve cluster leave and cluster readiness probing in general

dergraf:


apps/vmq_server/src/vmq_cluster.erl
```diff
@@ -169,12 +176,32 @@ check_ready([Node|Rest], Acc) ->
     ok = vmq_cluster_node_sup:ensure_cluster_node(Node),
     %% We should only say we're ready if we've established a
     %% connection to the remote node.
-    IsReady1 = IsReady andalso vmq_cluster_node_sup:is_reachable(Node),
+    S = vmq_cluster_node_sup:node_status(Node),
```
codeadict: Can this be named Status? I think will be more readable for contributors

dergraf: Sure! :+1:

apps/vmq_server/src/vmq_cluster.erl
```diff
@@ -169,12 +176,32 @@ check_ready([Node|Rest], Acc) ->
...
+        {false, {false, NetsplitDetectedCnt, NetsplitResolvedCnt}} ->
+            % Cluster was inconsistent, is still inconsistent
+            {false, NetsplitDetectedCnt, NetsplitResolvedCnt}
     end,
-    ets:insert(?VMQ_CLUSTER_STATUS, [{ready, ClusterReady}|Acc]).
+    ets:insert(?VMQ_CLUSTER_STATUS, [{ready, NewObj}|Acc]).
+
+-spec all_nodes_alive([{NodeName::atom(), IsReady::boolean()}]) -> boolean().
+all_nodes_alive(Nodes) ->
+    not lists:keymember(false, 2, Nodes).
```
larshesel: This seems kind of brittle to me, wouldn't it be better to replace with sth like `lists:all(fun({_, true, _}) -> true; (_) -> false end, Nodes)`.

PR #733, opened 2018-06-13
Gracefully convert new `vmq_msg` records to internal known format

larshesel:
Prepare VerneMQ to be able to handle:
- Cross cluster messages coming from a MQTTv5 enabled VerneMQ node
- Messages in the message store from a MQTTv5 node (downgrade)
- MQTTv5 Retained messages which are distributed via plumtree
- MQTTv5 Subscriber metadata which is distributed via plumtree


larshesel:
@dergraf  I think this is ready for review.

Issue #735, opened 2018-06-14
SSL connect timeout for constrained devices


mhball:
### Environment

- VerneMQ Version: 1.3.2
- OS: Ubuntu 16.04.4
- Erlang/OTP version (if building from source): N/A

### Expected behavior

The broker allows more time for constrained devices to process certificates for establishing SSL connections. Perhaps through some configuration option.

### Actual behaviour

We're trying to connect to the broker on a constrained embedded device (300MHz processor, ~512KB of RAM) using SSL.  Unfortunately, upon receiving the broker's server cert, it takes 6-8 seconds to process and verify; however, the broker is closing connection after ~5 seconds.

The SSL setup is working as expected when connecting with higher-powered machine, e.g. laptop.

larshesel:
Hi,

Thanks for opening the issue. That's a good point, this is unfortunately a hardcoded limit which is set to 5 seconds at the moment: https://github.com/erlio/vernemq/blob/master/apps/vmq_server/src/vmq_mqtt_fsm.erl#L22

We'll have to see if this is something we can tackle in the near future, but I can't make any promises at the moment.

larshesel:
Hmm, looking close at this, it's actually really simple to change so it's globally configurable. I'll create a PR with it and see if there's agreement to go with that approach.

codeadict:
perhaps something like `socket_close_timeout` in the config, that is actually super simple to implement.

mhball:
@larshesel thanks for the awesome quick response!
I pulled down and compiled your branch, and the set the configuration. It changed the timeout when we connect to port 1883 but for ssl to port 8883, it still defaulted to 5 seconds.

larshesel:
Ah, interesting. I didn't think about it before, but there might also be a timeout in the SSL stack - I'll check if that's the case.

codeadict:
isn't this related https://github.com/erlio/vernemq/blob/e305f8e8b1deabd5ddee487073563a2fa22feb4f/apps/vmq_server/src/vmq_server.app.src#L91

mhball:
Thanks for your help on this guys. It'd be great to get to the bottom of it.
As an interim solution we've set up a proxy to do SSL termination and forward to the broker on 1883

jmatt-cavli:
Any update on this? Or any roadmap?

ioolkos:
@jmatt-cavli thanks... no update, no roadmap. You have a use case, it seems?

codeadict:
[...]

PR #736, opened 2018-06-14
Make the connect timeout configurable

larshesel:
Fix for #735

### DIFF
--- a/apps/vmq_server/priv/vmq_server.schema
+++ b/apps/vmq_server/priv/vmq_server.schema
@@ -277,6 +277,13 @@
                                                                           hidden
                                                                          ]}.
 
+%% @doc specifies the connect timeout.
+{mapping, "connect_timeout", "vmq_server.connect_timeout", [
+                                                            {default, 5000},
+                                                            {datatype, integer},
+                                                            hidden
+                                                           ]}.
+
 %% @doc specifies the modules that should be evaluated when setting up the cowboy routes
 {mapping, "http_modules", "vmq_server.http_modules", [
                                                       {default, "[vmq_metrics_http,vmq_http_mgmt_api, vmq_status_http]"},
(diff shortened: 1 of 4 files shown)


apps/vmq_server/src/vmq_mqtt_fsm.erl
```diff
@@ -108,7 +108,8 @@ init(Peer, Opts) ->
                     },
     TraceFun = vmq_config:get_env(trace_fun, undefined),
 
-    TRef = vmq_mqtt_fsm_util:send_after(?CLOSE_AFTER, close_timeout),
+    ConnectTimeout = vmq_config:get_env(connect_timeout, 5000),
+    TRef = vmq_mqtt_fsm_util:send_after(ConnectTimeout, close_timeout),
```
codeadict: do you still need the `CLOSE_AFTER` macro? I think it can be removed to leave the code clean

apps/vmq_server/priv/vmq_server.schema
```diff
@@ -277,6 +277,13 @@
                                                                           hidden
                                                                          ]}.
 
+%% @doc specifies the connect timeout.
+{mapping, "connect_timeout", "vmq_server.connect_timeout", [
```
codeadict: isn't `socket_connect_timeout` more descriptive?

apps/vmq_server/src/vmq_mqtt_fsm.erl
larshesel: removed!

apps/vmq_server/priv/vmq_server.schema
larshesel: hmm, maybe sth like `mqtt_connect_timeout` would make it clear that it refers to MQTT connections?

codeadict: Yes this is actually better since it is explicit and there are multiple listeners and very generic can confuse users

dergraf:
@larshesel what is the status of this PR?

larshesel:
It doesn't solve the original issue. I'd say scrap it for now - so I'm closing it.

larshesel:
The original issue is still unresolved, of course.

Issue #738, opened 2018-06-20
Support Erlang/OTP 21


larshesel:
To get going on OTP 21 we need to tackle (at least) the below list of issues:

[ ] `eper` uses `get_stacktrace` which is deprecated
[ ] `clique` uses `get_stacktrace` which is deprecated
[ ] `vmq_ql` uses `get_stacktrace` which is deprecated
[ ] `vmq_server` uses `get_stacktrace` which is deprecated
[ ] `plumtree` is has `gen_fsm` (deprecated) overrides for opt 20, opt 21 needs to be added too
[ ] `rebar` needs to be upgraded to at least 3.5.2
[ ] use `logger` instead of `error_logger` (by changing or replacing the boot script?) otherwise we get the `no logger present` error.

larshesel:
Borrowed the macros for handling get_stacktrace from the rebar project: https://github.com/erlang/rebar3/pull/1773/files

PR #739, opened 2018-06-20
Support Erlang/OTP 21

larshesel:
See #738 for details.

This removes `eper` as a dependency, ignores warnings in `clique`, upgrades `rebar3` to version 3.5.3.

Still a few things to go. See the issue for details.


larshesel:
NOTE: upgrading to the new rebar3 again breaks loading external plugins: 
```
Can't enable plugin: vmq_allow_all_v5 due to {unknown_module,vmq_allow_all_v5}
```

We need to get to the bottom of this - this happened last time we upgraded rebar3 and we reverted back to fix this. Now we have to figure out what the problem is to be able to support OTP 21.

codeadict:
This is great, was the external plugin compiled with the same Rebar3?

larshesel:
I figured out what the problem was regarding rebar3/plugins. Sometime after rebar3 v 3.3.6 the mode was changed from `interactive` to `embedded`. In the embedded mode it's not possible to autoload modules (and hence plugin loading will fail). So we need to figure out how to stay in `interactive` mode. It seems it's the `CODE_LOADING_MODE` env variable which has to be set to `interactive`, but I haven't figured where to put this in `node_package`, `cuttlefish`/`rebar3_cuttlefish`,`rebar3` and `relx` to make it a default yet.

codeadict:
Seems like one can add `-mode interactive` to the `vm.args`, that will do the trick.

larshesel:
We have cuttlefish autogenerate the app.config and vm.args - I think the trick would be to get cuttlefish to put the correct stuff into the vm.args file... I got it to work if i do `export CODE_LOADING_MODE=interactive` before starting vernemq.

codeadict:
Yep the solution may be to submit a PR to cuttlefish

larshesel:
@dergraf can you bless this?

larshesel:
nevermind, need to fix some test-suites first...

PR #748, opened 2018-06-28
Remove cuttlefish warning about message_size_limit

codeadict:
The setting `message_size_limit` has been deprecated for long but still, when one runs `vernemq config generate` or starts the broker without a configuration file, it warns about this setting even when is not set. The setting has been deprecated for long and is not used anywhere, so we should remove it from the schema file.


codeadict:
Added this little PR @dergraf @larshesel

dergraf:
@larshesel I guess it is ok to remove that, but I am not sure for how many minor versions we wanted to keep a deprecation warning.

larshesel:
If we remove it entirely older config files will no longer parse which I think is a bit harsh as we should be backwards compatible ( I guess we broke semver to begin with when we made this setting do nothing...). Of course the setting is completely ignored, so if we could trigger the warning *only* if the user sets a value there then that would be better than removing it and break the old config files. Maybe only emit the warning if the value is `!= 0` ?

codeadict:
I agree with warning only if the setting is set, will update the PR later
today

codeadict:
Updated according to @larshesel suggestion, please review!

changelog.md
```diff
@@ -21,6 +21,7 @@
   VerneMQ configuration file. The default is `off`.
 - Fix issue with PUBREL frames retried after client reconnects (#762).
 - Refactor and cleanup retry mechanism.
+- Warn about `message_size_limit` only when it has been defined.
```
larshesel: For the readers of the changelog I think it would be nice to know this is about a deprecated setting so the context is cler, so maybe sth like  'Warn about deprecated  .... setting only when ...`?

codeadict:
Updated

larshesel:
Replaced by #774

PR #757, opened 2018-07-05
Validate shared subscription topic fixes #756

### LINKED ISSUE #756: Subscribing to $share/topic is killing the node
### Environment

- VerneMQ Version: 1.4.0-21c68522
- OS: Debian Stretch

### Expected behavior

Subscribe to a shared topic

### Actual behaviour

When i try to subscribe to `$share/topic` the verne node is going down.

1. with a test client try to subscribe to `$share/topic`
2. verne node is going offline right after you try to subscribe to the topic

jul-04-2018 23-17-06

dergraf:


ioolkos:
Does this send a 0x80 (Subscription failure) in the SUBACK? if not, would it make sense?

dergraf:
No, it validates it on the parser level... and terminates the connection.

Issue #762, opened 2018-07-10
Processing a received PUBCOMP after client reconnect crashes the session


dergraf:
### Expected behavior
A client dies, but it has unacked PUBREL frames. Once the client reconnects VerneMQ resends the PUBREL and expects a PUBCOMP. When VerneMQ processes the PUBCOMP it removes the PUBREL from the "waiting Acks".

### Actual behaviour
A client dies, but it has unacked PUBREL frames. Once the client reconnects VerneMQ resends the PUBREL and expects a PUBCOMP. When VerneMQ processes the PUBCOMP the session crashes.

This case should be covered in https://github.com/erlio/vernemq/blob/master/apps/vmq_server/test/vmq_publish_SUITE.erl#L177, however the testcase doesn't check that the session is actually still alive after sending the PUBCOMP.

The Crash log:

```
2018-07-10 09:48:32 =CRASH REPORT====
  crasher:
    initial call: vmq_ranch:init/5
    pid: <0.3415.0>
    registered_name: []
    exception error: {{case_clause,<<98,2,0,1>>},[{vmq_mqtt_fsm,connected,2,[{file,"/home/graf/Projects/vernemq/_build/test/lib/vmq_server/src/vmq_mqtt_fsm.erl"},{line,350}]},{vmq_mqtt_fsm,in,3,[{file,"/home/graf/Projects/vernemq/_build/test/lib/vmq_server/src/vmq_mqtt_fsm.erl"},{line,168}]},{vmq_mqtt_fsm,data_in,3,[{file,"/home/graf/Projects/vernemq/_build/test/lib/vmq_server/src/vmq_mqtt_fsm.erl"},{line,142}]},{vmq_ranch,handle_message,2,[{file,"/home/graf/Projects/vernemq/_build/test/lib/vmq_server/src/vmq_ranch.erl"},{line,174}]},{vmq_ranch,loop_,1,[{file,"/home/graf/Projects/vernemq/_build/test/lib/vmq_server/src/vmq_ranch.erl"},{line,117}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,247}]}]}
    ancestors: [<0.3385.0>,<0.3384.0>,ranch_sup,<0.3125.0>]
    message_queue_len: 0
    messages: []
    links: [<0.3385.0>,#Port<0.332260>]
[...]

Issue #750, opened 2018-06-29
Don't retry PUBREC frame


dergraf:
Temporarily stored PUBREC frames are retried if they aren't acknowledged with a PUBREL by the client.

PR #763, opened 2018-07-10
Fix QoS2 retry for reconnecting clients, and refactor and cleanup retry mechanism

dergraf:
- Store the *parsed* PUBREL inside the waiting acks instead of the
serialized binary.  This enables to reuse the same functionality for
handling retried PUBRELs, without caring if it is retried because of a
timeout or because of a reconnect. Cleanup places which relied on the
serialized binary.
- Fixing above enabled the possibility to implement a small performance
improvement where we could get rid of using `length()` on the outgoing
batch of frames for incrementing the `mqtt_publishes_sent` counter.
- Improved `vmq_publish_SUITE.erl` to test that a client responds to
PINGREQs before closing the Socket. This validates that 1st. the client
hasn't crashed in the meantime (e.g. because not being able to handle
the PUBCOMP, see bug #762), and 2nd that the last received frame on this
socket is the PINGRESP.
- The improvement above discovered an edgecase with a PUBREL frame that
has been retried too fast, indicating a bug in the retry mechanism. This
has been fixed with tagging the message IDs inside the retry queue. As a
result the retry mechanism can differentiate between retrying a PUBLISH
and a PUBREL. Without this fix a retried, already acked PUBLISH, would
have retried an unacked PUBREL right away.
- Removing several function clauses that were forgotten when fixing #750

### DIFF
--- a/apps/vmq_server/src/vmq_mqtt_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt_fsm.erl
@@ -312,7 +312,7 @@ connected(#mqtt_pubrec{message_id=MessageId}, State) ->
             PubRelFrame = #mqtt_pubrel{message_id=MessageId},
             _ = vmq_metrics:incr_mqtt_pubrel_sent(),
             {State#state{
-               retry_queue=set_retry(MessageId, RetryInterval, RetryQueue),
+               retry_queue=set_retry(pubrel, MessageId, RetryInterval, RetryQueue),
                waiting_acks=maps:update(MessageId, PubRelFrame, WAcks)},
             [PubRelFrame]};
         not_found ->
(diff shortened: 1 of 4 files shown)


larshesel:
Hi @dergraf 

I think this looks fine. I think there's another case which isn't covered: we don't persist the pubrel (deliver_bin) messages in `vmq_queue:maybe_offline_store/3`, but we rely in the binary pubrel message to be passed to the new session and handle it in `vmq_mqtt_fsm:handle_messages/5`. This works as long as the offline queue is held in memory, but not if it is read back from disk (in case of a node restart). IIUIC, of course :)

Btw, the remaining `deliver_bin` cases seem to me to be unnecessary as we only use this to mark it as a pubrel frame - and since we parse the binary before we use it in the fsm I suppose we could just store the msg unserialized? But I have a feeling I'm missing something here.

PR #767, opened 2018-07-17
Suppress lwt on session take over

larshesel:
This PR implements a new setting which can suppress LWT messages in case of a session takeover by setting the `suppress_lwt_on_session_takeover=on`. The new setting is hidden (not visible by default in the vernemq.conf file).

To implement this d075ff3 - 'Add reasons to queue cleanup and session disconnects' was backported/cherry picked from the mqtt5-preview branch as reason codes had already been generalized there.

This PR also introduces the `def_opts` map member to the `vmq_mqtt_fsm` state record, which is meant to hold non-fast-path settings. The idea is that if we store all settings directly in the record we'll have a memory overhead for all sessions, but if instead only non-default values are stored in the map, we don't have any overhead for default values. We can move other settings into this map later if needed.

This PR also refactors and simplifies a bit the `vmq_mqtt_fsm:terminate/2` and `vmq_mqtt_fsm:maybe_publish_last_will/1` functions, as well as introducing a function `vmq_mqtt_fsm:terminate_reason/1` which maps reasons to `normal` to make the ranch process shut down properly for normal shutdown reasons.

TODO: see how this merges with the MQTT5 branch.

### DIFF
--- a/apps/vmq_server/priv/vmq_server.schema
+++ b/apps/vmq_server/priv/vmq_server.schema
@@ -1449,6 +1449,15 @@
                                                    hidden
                                                   ]}.
 
+%% @doc Suppress the Last Will and Testament message if an existing
+%% session already exists for the connecting client.
+{mapping, "suppress_lwt_on_session_takeover",
+ "vmq_server.suppress_lwt_on_session_takeover", [
+                                                 {default, off},
+                                                 {datatype, flag},
+                                                 hidden
+                                                ]}.
+
 %% @doc plugins.<plugin> enables/disables a plugin.
 %%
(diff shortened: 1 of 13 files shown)


ioolkos:
Note: it's `suppress_lwt_on_session_takeover=on|off` not `suppress_lwt_on_session_takeover=true|false`

larshesel:
No, there's a lot of stuff we could log and improve as we can now differentiate between the different disconnect reasons. This PR just addresses the session takeover case. But let's cook something up with this (perhaps together with the MQTT5 branch?)

PR #770, opened 2018-07-21
add CockroachDB auth plugin

cmnstmntmn:
- all credits goes to @knz
- tests should be reviewed

### DIFF
--- a/apps/vmq_diversity/priv/auth/crdb.lua
+++ b/apps/vmq_diversity/priv/auth/crdb.lua
@@ -1,0 +1,111 @@
+-- Copyright 2018 Erlio GmbH Basel Switzerland (http://erl.io)
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


ioolkos:
@cmnstmntmn, @knz cool stuff, thanks! :)

dergraf:
@cmnstmntmn had a very quick look at the Lua code... one question is it basically the Postgres code but using SHA hashes instead of the PG crypto stuff?

cmnstmntmn:
@dergraf, indeed

dergraf:
While this could be merged that way, I'd prefer a more generic solution where we could e.g. change the hashing/password mechanism, this could be useful for the other database plugins as well, especially those that currently live with the potential bcrypt bottleneck... @larshesel any thoughts on this?

larshesel:
I think that sounds like a good approach.

larshesel:
As we'd rather have a generic approach I'll close this for now.

PR #776, opened 2018-07-26
Rename mqtt5 hooks

larshesel:
As discussed a while back I think it's confusing to call the new hooks `_v1` introduced in the mqtt5 fsm. They are simply new hooks needed to support the mqtt version 5 protocol and not new versions of the old hooks with the similar names. With this change we can later add new hook versions without it being too confusing...

This depends on the same change in the `vernemq_dev` repo: https://github.com/larshesel/vernemq_dev/pull/1.

Are the `_m5` postfix acceptable for you?


larshesel:
great, merged.

PR #777, opened 2018-07-27
Use git hashes in non-tagged package names (nightly builds)

larshesel:
This makes it easier to reason about nightly bulds and the features
they contain.

Hopefully this doesn't break our download page as this means package names now can contain one more `-` (now: `vernemq_1.4.0-10-gde1b1f5-1_amd64.deb`, before: `vernemq_1.4.0-b994c136-1_amd64.db`). Looking over the code I don't think this will be a problem, though.


larshesel:
I guess we'll see quite soon if something breaks (as soon as this is merged and a nightly has been built ;) )

PR #780, opened 2018-07-30
Merge generalized metrics to mqtt5 preview

larshesel:
Wanna prepare metrics for the mqtt5 fsm so I needed to have the generalized metrics logic merged into the mqtt5-preview branch


larshesel:
redid the merge commit and the tests passed.

PR #795, opened 2018-08-13
Support max_packet_size and request_problem_information CONNECT properties

dergraf:
Supersedes PR #710 
Same TODOs remain:
This PR adds the support for the Maximum Packet Size CONNECT property.

- [x] Parse CONNECT property ans pass it into the MQTT session (currently in the process registry)
- [x] Use the property to strip User Property if the packet size exceeds the Maximum Packet Size
- [x] Test stripping of User Property
- [x] Use the property to strip Reason Strings if the packet size exceeds the Maximum Packet Size
- [ ] Test stripping of Reason Strings (we currently don't have Reason Strings, how to add them?)
- [ ] If packet is still too big after stripping cleanup state (it's actually not very clear how to do this e.g. if you are in the middle of a QoS 2 flow and e.g. we can't send the PUBREL because it is too big.)

- [x] Parse Request Problem Information CONNECT property and pass it into the MQTT session (currently in the process registry)
- [x] Use the Request Problem Information Property to strip User Property in case of error (SUBACK/UNSUBACK currently missing)
- [x] Use the Request Problem Information Property to strip Reason Strings in case of error (SUBACK/UNSUBACK currently missing)

### DIFF
--- a/apps/vmq_server/test/vmq_publish_SUITE.erl
+++ b/apps/vmq_server/test/vmq_publish_SUITE.erl
@@ -84,7 +84,8 @@ groups() ->
                    message_expiry,
                    publish_c2b_topic_alias,
                    publish_b2c_topic_alias,
-                   forward_properties
+                   forward_properties,
+                   max_packet_size
                    | V4V5Tests] }
     ].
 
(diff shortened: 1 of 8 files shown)


apps/vmq_server/test/vmq_publish_SUITE.erl
```diff
@@ -833,6 +834,34 @@ forward_properties(_Config) ->
...
+    PubConnack = packetv5:gen_connack(0, ?M5_CONNACK_ACCEPT, #{}),
+    {ok, PubSocket} = packetv5:do_client_connect(PubConnect, PubConnack, []),
+    ok = gen_tcp:send(PubSocket, Pub),
+    ok = packetv5:expect_frame(SubSocket, ReducedPub),
+    ok = gen_tcp:close(PubSocket),
+    ok = gen_tcp:close(SubSocket),
+    disable_on_publish(),
+    disable_on_subscribe(),
+    ok.
+
```
larshesel: It looks like dropping messages hasn't been implemented? `Where a Packet is too large to send, the Server MUST discard it without sending it and then behave as if
it had completed sending that Application Message [MQTT-3.1.2-25].`

larshesel:
> If packet is still too big after stripping cleanup state (it's actually not very clear how to do this e.g. if you are in the middle of a QoS 2 flow and e.g. we can't send the PUBREL because it is too big.)

Regarding the above, then I interpret `MQTT-3.1.2.-25` as referring to the application messages. So if you can't send a PUBREL then you should just behave as had you completed the entire QoS2 machinery: IE just forget all about that msg id and cleanup what needs to be cleaned up.

ckrey:
The spec says you MUST NOT sent "Reason String" or any "User Property" as properties if you would create a package larger than the maximum message size.

The basic PUBREL packet is only 5 octets (fixed header, remaining length, packet identefier MSB and LSB, reason code). 

If the client specified the maximum message size smaller than 5, we are in trouble anyhow. Unfortunately the standard does not specify a lower limit to the maximum message size.

I guess the same is true for PUBACK, PUBREC and PUBCOMP and likewise for SUBACK, UNSUBACK, PINGRESP, DISCONNECT, CONNACK, and AUTH.

larshesel:
Awesome point @ckrey . Btw, do you happen to know which use-case brought this stripping feature into the spec? To me it seems a rather complex part of the spec and I haven't yet figured out where this could be super important.

ckrey:
@larshesel no I don't. I would have liked to see a minimum packet size which is required by all client and server implementations which avoids stripped down messages to be dropped.

From my point of view stripping user properties in PUBLISH packets may make sense because a subscribing client doesn't necessarily know which and how many User Properties are published on the other end.

The delivery of Reason String and User Properties for packages other than PUBLISH, DISCONNECT, or CONNACK is controlled by the client by setting the Request Problem Information property in the CONNECT packet (see 3.1.2.11.7).

PR #800, opened 2018-08-16
Add mqttv5 support to existing plugins

larshesel:
Builds on top of #799 

- [x] `vmq_passwd`
- [x] `vmq_acl`
- [x] `vmq_webhooks`
- [ ] `vmq_diversity`
- [ ] `vmq_bridge`


larshesel:
The webhook endpoint communication look like this:

Sent: 

```json
{
  "username": "test-user",
  "mountpoint": "mountpoint",
  "client_id": "allowed-subscriber-id",
  "properties": {
    "authentication_data": "QVVUSF9EQVRBMA==",
    "authentication_method": "AUTH_METHOD"
  }
}
```

Received:
```json
{
  "modifiers": {
    "properties": {
      "authentication_data": "QVVUSF9EQVRBMQ==",
      "authentication_method": "AUTH_METHOD"
    },
    "reason_code": 0
  },
  "result": "ok"
}
```

auth data is *always* base64 encoded as the type of that property is binary. 

Another example (which deviates from the older hooks) is `auth_on_subscribe`: 

Sent:
```json
{
  "username": "test-user",
  "mountpoint": "mountpoint",
  "client_id": "changed-subscriber-id",
  "topics": [
    {
      "topic": "test/topic",
      "qos": 1
    }
  ],
  "properties": {}
}
```

Received:
```json
{
  "modifiers": {
    "topics": [
      {
        "qos": 2,
        "topic": "rewritten/topic"
      },
      {
        "qos": 135,
        "topic": "forbidden/topic"
      }
    ]
  },
  "result": "ok"
}
```

Note how the result of the subscribe is wrapped in a modifiers object, which wasn't the case for the `auth_on_subscribe` hook. The reason is that we can now also return properties (not shown). This change unifies the handling of the responses of all the `all_till_ok` mqtt5 hooks as they're now all on the same form.

larshesel:
rebased

larshesel:
What is also missing a proper tests for all the properties, for example on `auth_on_publish`... but I suppose that could wait until later..

Issue #588, opened 2018-01-22
Lua scripting docs don't mention it's critical to call http.body after making an http request


originsmike:
This is just a doc issue with [the lua scripting page](https://vernemq.com/docs/plugindevelopment/luaplugins.html#http-and-json-client-libraries). Just want to save the next person a few hours of debugging :)

I was tipped off by https://github.com/erlio/vernemq/issues/341 ; basically, if you make http requests without calling http.body on a non-false response, you're leaking some sort of HTTP sessions inside hackney. The tricky part is that when the sessions are exhausted, the error you see is connection_timeout, which it takes the full 8 seconds to report, so it looks like a genuine timeout. The timeout error also tears down the current client's MQTT session, along with any other sessions that are unfortunate enough to currently be waiting their turn to execute a hook.

Fixing the docs would help. A better fix would be to change the error message so it's not a red herring and tells you what you're doing wrong.

I don't call this a bug because golang's HTTP client library has the same failure mode. See https://github.com/golang/go/issues/17166

ioolkos:
@originsmike thanks a lot for reporting!

not sure I understand it fully on first read, but no worries, we'll try to fix the doc. (cc @larshesel @dergraf )
We really appreciate your effort and letting us and future users know about it!  :thumbsup:

larshesel:
HI @originsmike I took a look at how we handle the http connections using hackney and hackney will keep the connection from going back to the pool until the user has read everything it can using the reference returned from the request. So as you correctly found out, not calling the body function will lead to a complete exhaustion of the connection pool and everything grinds to a standstill and timeouts happen.

It's easy to document that one *must* call http.body. Changing the error is harder (as it comes from hackney). Another possibility is to tell hackney to return the actual body instead of a reference which could be later read (this would free up the connection immediately AFAICT) using the `with_body` option (https://github.com/benoitc/hackney/#read-the-body). Then users would not have to call http.body at all as it would be returned directly - however that would mean a breaking change to the lua interface.

I'm not sure what the right approach would be, but I'll for sure make an update to the docs for now.

originsmike:
@larshesel could you use the `with_body` option and hang the response body off the returned response object, making `http.body` just extract the already-fetched response body? If garbage collection in Lua is smart enough, that wouldn't require an API change.

[...]

PR #802, opened 2018-08-17
Fetch http body directly to avoid drying up the connection pool

larshesel:
This handles the issue in #588 - if this is approved I'll adjust the docs accordingly.


apps/vmq_diversity/src/vmq_diversity_http.erl
```diff
@@ -44,29 +44,27 @@ request(Method, [BPoolId, Url|Rest0] = As, St) when is_binary(Url) ->
...
                      {headers, RespHeaders},
-                     {ref, BBClientRef}],
+                     %% We not longer have a reference, but we store
+                     %% the body in the reference to not break
+                     %% existing scripts. Earlier a reference was
+                     %% returned, but if the user didn't call body on
+                     %% it, the underlying connection would not be
+                     %% returned to the connection pool, drying up the
+                     %% pool.
+                     {ref, Body}],
```
dergraf: yeah, nice workaround

PR #807, opened 2018-08-24
Don't timeout when calling a Lua script function.

### LINKED ISSUE #589: Lua plugin timeout (5s) is shorter than http request timeout (8s)
### Environment

- VerneMQ Version: 1.2.3
- OS: ubuntu 16.04
- Erlang/OTP version (if building from source): n/a
- Running inside Docker 17.09, ports exposed in host mode

### Expected behavior

I've got a lua script where I'm making a call to http.post about 1% of the time in the on_publish hook. If the call fails due to a connection timeout, I'd expect to be able to ignore the error, and not have it affect other sessions that were unfortunate enough to happen to be waiting to execute a script hook at that moment.

### Actual behaviour

If the HTTP call fails with a timeout, it takes 8 seconds to fail, which isn't terrible but it's longer than the 5 second timeout set for script execution in general. So while the http timeout itself is just treated as a normal failure (and that hook continues executing), the consequence for exceeding the 5 second timeout is that your hook crashes and the sandboxing code responds by tearing down its MQTT session.

Any other hooks that were unfortunate enough to get stuck in line behind the hook that timed out, also exceed the 5 second threshold and have their MQTT sessions torn down, regardless of whether they were going to make http requests or not. You can mitigate this by setting num_states to something higher than 1 though.

Logs:

```
2018-01-22 07:13:16.542 [error] <0.6696.0>@vmq_diversity_script:call_function:64 can't call into Lua sandbox for function on_publish due to {timeout,{gen_server,call,[<0.767.0>,{call_function,on_publish,[{username,<<"e****">>},{mountpoint,<<>>},{client_id,<<"client1****">>},{qos,1},{topic,<<"l****/i****">>},{payload,<<"payload1****">>},{retain,false}]}]}}
2018-01-22 07:13:17.844 [error] <0.6395.0>@vmq_diversity_script:call_function:64 can't call into Lua sandbox for function on_publish due to {timeout,{gen_server,call,[<0.767.0>,{call_function,on_publish,[{username,<<"s****">>},{mountpoint,<<>>},{client_id,<<"client2****">>},{qos,1},{topic,<<"w****/c****">>},{payload,<<"payload2****">>},{retain,true}]}]}}
2018-01-22 07:13:18.442 [error] <0.6578.0>@vmq_diversity_script:call_function:64 can't call into Lua sandbox for function on_publish due to {timeout,{gen_server,call,[<0.767.0>,{call_function,on_publish,[{username,<<"e****">>},{mountpoint,<<>>},{client_id,<<"client3****">>},{qos,1},{topic,<<"l****/h****">>},{payload,<<"payload3****">>},{retain,false}]}]}}
2018-01-22 07:13:19.543 [error] <0.764.0>@vmq_diversity_http:request:57 http request failure for post <<"http://my-http-server/events">> due to connect_timeout
2018-01-22 07:13:19.544 [error] <0.764.0>@vmq_diversity_lager:error:36 [[[my-script.lua]]] http.post returned false
2018-01-22 07:13:19.545 [debug] <0.6696.0>@vmq_mqtt_fsm:connected:432 stopped connected session, due to unexpected frame type {#Ref<0.0.1.115572>,undefined}
2018-01-22 07:13:19.546 [debug] <0.6578.0>@vmq_mqtt_fsm:connected:432 stopped connected session, due to unexpected frame type {#Ref<0.0.1.115605>,undefined}
2018-01-22 07:13:19.546 [warning] <0.6578.0>@vmq_ranch:teardown:131 session stopped abnormally due to '{error,unexpected_message,{#Ref<0.0.1.115605>,undefined}}'
2018-01-22 07:13:19.547 [warning] <0.6696.0>@vmq_ranch:teardown:131 session stopped abnormally due to '{error,unexpected_message,{#Ref<0.0.1.115572>,undefined}}'
2018-01-22 07:13:19.548 [debug] <0.6699.0>@vmq_queue:state_change:913 transition from online --> offline because of 'DOWN'
2018-01-22 07:13:19.549 [debug] <0.6395.0>@vmq_mqtt_fsm:connected:432 stopped connected session, due to unexpected frame type {#Ref<0.0.1.115584>,undefined}
2018-01-22 07:13:19.550 [warning] <0.6395.0>@vmq_ranch:teardown:131 session stopped abnormally due to '{error,unexpected_message,{#Ref<0.0.1.115584>,undefined}}'
2018-01-22 07:13:19.551 [debug] <0.695.0>@vmq_queue:state_change:913 transition from online --> offline because of 'DOWN'
2018-01-22 07:13:19.559 [debug] <0.695.0>@vmq_queue:state_change:913 transition from online --> offline because of 'DOWN'
```

In this case, only client1's on_publish hook made the http.post call; the others didn't, but they still got unlucky.

### Background & stuff

Basically, this issue is about trying to limit the collateral damage if one of the back end services happens to be unhealthy.

(what I'm really trying to do is somehow get a feed of connecting clients' IP addresses into Kafka, and I happen to already have an HTTP service that forwards events to Kafka so I'm just using that. This IP address feed is nice to have but isn't mission-critical, so if the HTTP service isn't healthy, I don't want it to completely kill throughput for other MQTT clients)

To make this bulletproof, I can see a few solutions:

1. Write an Erlang plugin that uses the official Kafka client, and has strict timeouts
2. Deploy the HTTP->Kafka service locally with VerneMQ, and make sure it's well-behaved
3. If http calls start failing by returning false, detect this and don't make any calls for a while (save this state in ETS or something)

But my hunch is that this timeout issue affects database calls in addition to http requests, so solutions 2 and 3 above are only workarounds. (seems like [mysql's default is 8 seconds](https://github.com/Eonblast/Emysql/blob/c7e2103f8b737667f0128802e8de2f0d5ed2fa5c/include/emysql.hrl#L114) as well)

Also, if you're trying to do something like https://github.com/erlio/vernemq/issues/197 (which I'll probably do eventually) then this timeout thing becomes a lot more important.

dergraf:
Fixes #589

### DIFF
--- a/apps/vmq_diversity/src/vmq_diversity_script.erl
+++ b/apps/vmq_diversity/src/vmq_diversity_script.erl
@@ -59,7 +59,7 @@ call_function(Pid, Function, Args) ->
     %% error in the plugin code can crash session/queue. As the Lua
     %% support should provide a sandboxed environment we saveguard
     %% calls into the Lua environment.
-    case catch gen_server:call(Pid, {call_function, Function, Args}) of
+    case catch gen_server:call(Pid, {call_function, Function, Args}, infinity) of
         {'EXIT', Reason} ->
             lager:error("can't call into Lua sandbox for function ~p due to ~p", [Function, Reason]),
             error;
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Fix issue when calling a function in a Lua script that requires more time to complete than the default `gen_server` timeout (#589). 
+
 ## VerneMQ 1.5.0
 
 - Fix issue in the bridge preventing it from reconnecting after a connection


dergraf:
this was actually already the case as the Lua State runs in another gen_server... we may add some warnings for long running scripts in the future.

PR #818, opened 2018-09-03
Prevent publish topics prefixed with $

dergraf:


dergraf:
This PR is for mqtt5-preview... the changelog was added to the PR #817 which addresses the MQTT 4.

PR #820, opened 2018-09-03
Prepare for merge

larshesel:
A few of the pre-merge items to cross off of the list...

### DIFF
--- a/apps/vmq_commons/include/vmq_types_mqtt5.hrl
+++ b/apps/vmq_commons/include/vmq_types_mqtt5.hrl
@@ -239,7 +239,7 @@
 
 -record(mqtt5_unsubscribe, {
           message_id        :: msg_id(),
-          topics=[]         :: [topic()],
+          topics=[]         :: nonempty_list(topic()),
           properties=#{}    :: mqtt5_unsub_props()
          }).
 -type mqtt5_unsubscribe()   :: #mqtt5_unsubscribe{}.
--- a/apps/vmq_commons/src/vmq_parser_mqtt5.erl
+++ b/apps/vmq_commons/src/vmq_parser_mqtt5.erl
@@ -25,6 +25,8 @@
          rc2rcn/1
         ]).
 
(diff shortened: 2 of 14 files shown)


larshesel:
Hey @dergraf , this is interesting. Notice a42f86c and that I reverted it in 8f725d3. Dialyzer correctly complains:

```
apps/vmq_server/src/vmq_metadata.erl
  32: Call to missing or unexported function vmq_plugin_mgr:disable_plugin/2
```

As there *is* no arity two disable_plugin function in `vmq_plugin_mgr`. But with the fix a test like 

```
rebar3 ct --suite ./apps/vmq_server/test/vmq_clean_session_SUITE.erl --group mqttv4 --case clean_session_qos1_test
```

Just hangs in the suite teardown which calls `vmq_server:stop()` - which ends up calling `vmq_plugin_mgr:disable_plugin`. The test passes when the undefined function is used, but hangs when the correct one is used... got an idea what the problem could be? I reverted the commit with the fix to be able to run the test suites for now

larshesel:
Hey @dergraf I'm done with this - please review. I investigated the dialyzer error about `vmq_plugin_mgr:disable_plugin/2` when it's fixed, `vmq_plugin_mgr:disable_plugin(vmq_plumtree)` simply doesn't return. Maybe you could take a look at that?

dergraf:
Analyzed the case. So the root of the problem is a deadlock in the OTP application controller when calling `vmq_server:stop()`.

```
vmq_server:stop(
    application:stop(vmq_server,
        vmq_server_app:stop(
             vmq_metadata:stop(
                  vmq_plugin_mgr:stop(vmq_plumtree
                      application:stop(vmq_plumtree ===>>>BOOM!
```
As `application:stop` is a blocking call, we can't stop another application in the terminate path of an application.

Simple fix would be to run `vmq_metadata:stop` inside a own process.

dergraf:
I've a PR for the issue above in #829 
So I'd say revert the related changes here... and merge this PR.

Issue #823, opened 2018-09-06
Array of topics is reversed in auth_on_subscribe webhook leading to qos mismatch


doubliez:
### Environment

- VerneMQ Version: 1.5.0
- OS: Docker/Kubernetes
- Erlang/OTP version (if building from source):

### Expected behavior

Using the MQTT.js client (https://www.npmjs.com/package/mqtt), I'm subscribing to some topics, let's say A, B and C, with qos=1:

```javascript
client.subscribe(['A', 'B', 'C'], { qos: 1 }, (err, granted) => {
    console.log('granted', granted);
});
```

I have implemented the auth_on_subscribe webhook that I use to authorize access to certain topics only. In the response I return a "topics" array identical to the array I get in the request except for the qos which is rewritten to 128 for those topics that are not authorized.

Let's say the client is allowed to subscribe to topic A, but not to topic B and C.

The above code in the MQTT client **should** log:

```
granted [ { topic: 'A', qos: 1 },
  { topic: 'B', qos: 128 },
  { topic: 'C', qos: 128 } ]
```

### Actual behaviour

The MQTT client logs:

```
granted [ { topic: 'A', qos: 128 },
  { topic: 'B', qos: 128 },
  { topic: 'C', qos: 1 } ]
```

After investigating the issue, I noticed that the array of topics I receive in the auth_on_subscribe webhook is reversed compared to what the client is sending, and that if I reverse the array before returning it in the response I send for auth_on_subscribe, the client logs the expected "granted" array.

This doesn't seem to affect which topics actually get subscribed in VerneMQ, _i.e._ only topic A would be subscribed (as expected).

So I have a few questions:
1. Why is VerneMQ sending a reversed "topics" array in the payload of auth_on_subscribe compared to what the client is sending?
2. From what I understand, the SUBACK message sent to the client must have the return codes (qos values) in the same order as the topics sent by the client, and must have the same number of return codes as the number of topics sent by the client. However it seems that the SUBACK message sent by VerneMQ is crafted from whatever is sent in the response of auth_on_subscribe. This is maybe not an issue, but it should be clearly documented that when sending topics in the response of auth_on_subscribe, all topics should be specified, not only those to rewrite. And they should be sent in the same order as sent by the client, which is made difficult given the previous point (reversed array). Is this correct?

ioolkos:
@doubliez thanks for this catch...and thanks for your effort to make this into a detailled report. Very much appreciated!

I'm just gonna say this is a bug, clearly.

[...]

PR #826, opened 2018-09-10
Fix ordering bug in webhook subscribe topic rewrites (#823)

larshesel:
This is a fix for #823 - but I think we should wait to merge this until after we've merged the mqtt5 branch #814


larshesel:
@dergraf Had to rewrite this after the rebase - still ok to merge?

PR #832, opened 2018-09-11
Run shared subscriptions tests on mqttv4 and mqttv5

larshesel:
Make the shared subscription tests run on both mqttv4 and mqttv5...


larshesel:
btw didn't add a changelog to this as I consider this purely internal stuff.

PR #835, opened 2018-09-13
Remove vmq_allow_all_v5 plugin

larshesel:
Forgot to remove this plugin... shouldn't be part of the master branch.


larshesel:
I just went ahead and merged this - I really don't want to have this enabled by default in master - if you disagree we can revert this change.

dergraf:
all good!

PR #838, opened 2018-09-14
Add MQTT 5.0 support to MySQL, Redis, PostgreSQL, MongoDB auth

larshesel:
This adds support for using the `vmq_diversity` database authentication drivers with MQTT 5.0 clients.

Note we now normalize the output of the `auth_on_subscribe` lua hook so we only have to handle the canonical format in the `vmq_plugin_util:check_modifiers/2` function - before we handled whatever came out of the lua hook, see https://github.com/erlio/vernemq/compare/master...larshesel:add-mqttv5-support-to-existing-plugins?expand=1#diff-f6d817c136d8a74d88f87d39403c3061L296

Note, the rest of the MQTT 5.0 hooks haven't been exposed to Lua yet.

### DIFF
--- a/apps/vmq_diversity/src/vmq_diversity_plugin.erl
+++ b/apps/vmq_diversity/src/vmq_diversity_plugin.erl
@@ -28,6 +28,10 @@
 -behaviour(on_client_offline_hook).
 -behaviour(on_client_gone_hook).
 
+-behaviour(auth_on_register_m5_hook).
+-behaviour(auth_on_publish_m5_hook).
+-behaviour(auth_on_subscribe_m5_hook).
+
 -export([auth_on_register/5,
          auth_on_publish/6,
          auth_on_subscribe/3,
@@ -39,7 +43,10 @@
          on_offline_message/5,
          on_client_wakeup/1,
          on_client_offline/1,
-         on_client_gone/1]).
+         on_client_gone/1,
+         auth_on_register_m5/6,
+         auth_on_publish_m5/7,
+         auth_on_subscribe_m5/4]).
 
 
 %% API functions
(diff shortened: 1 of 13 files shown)


apps/vmq_diversity/src/vmq_diversity_plugin.erl
```diff
@@ -28,6 +28,8 @@
 -behaviour(on_client_offline_hook).
 -behaviour(on_client_gone_hook).
 
+-behaviour(auth_on_register_m5_hook).
```
dergraf: do we need the other behaviours too?

larshesel: yes, good catch

larshesel: fixed

changelog.md
```diff
@@ -39,11 +39,13 @@
   example `listener.tcp.allowed_protocol_versions=3,4,5` would enable MQTT
   version 3.1, 3.1.1 and 5.0 on the TCP listener.
 
-  MQTTv5 support has also been added to the `vmq_passwd`, `vmq_acl`,
-  `vmq_webhooks` plugins.
+  MQTTv5 support has been added to the `vmq_passwd`, `vmq_acl`, `vmq_webhooks`
+  plugins as well as the parts of the `vmq_diversity` plugin to support the
+  MySQL, PostgreSQL, Redis and MongoDB authentication and authorization
+  mechanisms. The Lua scripting lange provided by the `vmq_diversity` plugin
```
dergraf: `lange` ? typo?

larshesel:
@dergraf I had to do a small fix to make the tests pass - can you take a look - basically the output of the cache for subscriptions was still in lua format and would therefor fail the modifier check pass.

larshesel:
ah, damn, looks like I broke sth in the webhooks suite - will take a look at that on my way home...

larshesel:
Fixed the webhooks suite.... it was actually returning subscription modifiers like `vmq_diversity` used to: `[[Topic,Qos]]`

dergraf:
All good for me!

PR #842, opened 2018-09-17
Add missing counters

larshesel:
Fix some missing metrics reported by @ckrey (thanks!) as well as one which was simply wrong (sending a disconnect where a connack with error code should've been sent...)

### DIFF
--- a/apps/vmq_server/src/vmq_metrics.erl
+++ b/apps/vmq_server/src/vmq_metrics.erl
@@ -596,7 +596,8 @@ metric_values() ->
 
 internal_defs() ->
     flatten([counter_entries_def(), mqtt4_connack_sent_def(),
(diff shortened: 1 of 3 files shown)


larshesel:
btw I didn't add a changelog item to as this is a bugfix for a not yet released feature.

apps/vmq_server/src/vmq_metrics.erl
```diff
@@ -698,7 +699,7 @@ rcn_to_str(RNC) ->
     %% TODO: replace this with a real textual representation
     atom_to_list(RNC).
 
-mqtt5_disconnect_def() ->
+mqtt5_disconnect_recv_def() ->
```
dergraf: naming... `mqtt5_disconnect_received_def` like `mqtt5_puback_received_def` ?

PR #843, opened 2018-09-17
Added Solaris build fixes

siepkes:
This patch makes VerneMQ buildable on SmartOS / Illumos / Solaris. 

The 1.0.0 to 1.0.2 bump in the bcrypt library adds Solaris support the the bcrypt library. It doesn't change anything else in the bcrypt library. It's a 2 minor version jump because the first release did not fix all Solaris build issues in the bcrypt library (as you see here: https://github.com/erlangpack/bcrypt ).

I think it will be best to hold off merging this PR until this PR: https://github.com/erlio/mzmetrics/pull/2 is merged first. I can then update the `mzmetrics` version in `rebar.lock` and update this PR to fix all build issues.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -70,6 +70,7 @@
 - Fix ordering bug in webhook subscribe topic authentication and topic rewrites
   (#823)
 - Fix issue when terminating the `vmq_server` application (#828).
+- Make VerneMQ build on SmartOS / Illumos / Solaris.
 
 ## VerneMQ 1.5.0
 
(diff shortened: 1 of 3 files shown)


larshesel:
Oh, could you add an item to the changelog file (at the bottom of the list of unreleased bullet points, just above the 1.5.0 heading) like:

```
- Make VerneMQ build on SmartOS / Illumos / Solaris.
```

siepkes:
Rebased and changelog entry added! Thanks for the build fix!

I just noticed I removed the `bcrypt` builder override because it was no longer necessary. Though iirc removing that override is not necessary to make it work on Solaris. Do you want me to put it back in or leave it out?

larshesel:
Thanks for the contribution. The issue with the port compiler was that a bad version was published and rebar3 isn't currently able to automatically lock plugins... since the problem seem to have been resolved it's ok to use a newer version. Will tackle pegging/locking plugins later.

Thanks again!

PR #850, opened 2018-09-20
Update m5 hooks

larshesel:
This PR changes the hooks according to erlio/vernemq_dev#6: Properties are no longer passed directly from hook implementations, but are created in the FSM as needed. I'm sure we can come up with something nicer and easier on the eye than my `maps:fold` functions for picking the modifiers which go into the properties.

Note the `sg_policy()` type was moved to the erlio/vernemq_dev repo and renamed to `shared_sub_policy()`.

### DIFF
--- a/apps/vmq_server/src/vmq_mqtt5_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt5_fsm.erl
@@ -83,7 +83,7 @@
           max_client_id_size=100            :: non_neg_integer(),
 
           %% changeable by auth_on_register
-          shared_subscription_policy=prefer_local :: sg_policy(),
+          shared_subscription_policy=prefer_local :: shared_sub_policy(),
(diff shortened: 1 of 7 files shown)


apps/vmq_server/src/vmq_mqtt5_fsm.erl
```diff
@@ -1049,7 +1098,7 @@ unsubscribe(User, SubscriberId, Topics0, Props0, UnsubFun) ->
 
 -spec auth_on_publish(username(), subscriber_id(), msg(),
                       fun((msg(), list()) -> {ok, msg()} | {error, any()})
-                        ) -> {ok, msg()} | {error, atom() | list()}.
+                        ) -> {ok, msg()} | {error, atom() | {atom(),map()}}.
```
dergraf: `{atom(), map()}` could we narrow it to `{ok |· error, map()}` ?

PR #881, opened 2018-10-05
Optimize subscribe unsubscribe fanout

larshesel:


larshesel:
I think it might be useful if we want to improve future improvements? In fact I think we should have more of these.

Note these benchmarks aren't run when just running `rebar3 ct` you have to specifically run them.

PR #888, opened 2018-10-09
Rebase Master

dergraf:


apps/vmq_swc/priv/vmq_swc.schema
```diff
@@ -0,0 +1,12 @@
+%% @doc Set the path to a password file.
```
codeadict: Seems like this comment was kept fro something else?

PR #901, opened 2018-10-17
Fixes retry bug in gen_mqtt_client

dergraf:
This commit fixes several bugs in `gen_mqtt_client` that could have been
observed when using the vmq_bridge with automatic reconnect configured.

What has been fixed:
- When automatic reconnect and `clean_session=true` is configured the
client didn't remove the waiting acks and unacked frames on disconnect.
As a consequence the unacked frames were retried and potentially bleeded
into the new 'clean' session.
- If `clean_session=false` the retried frames were lost if the client
was in a different state then `connected` e.g. while reconnecting.
- If `clean_session=false` the frames were retried when their timer
fired. However as it is a new connection the old timers don't matter
anymore and those frames should be retried right after sending out the
connack.
- When receiving a MQTT Disconnect frame the client didn't close the
socket but created a new one on reconnect. Unless the remote end
properly closed the old socket, the old socket was still owned by the
client process. Moreover, any TCP message (data, error, close) on such
an old socket would have triggered a `handle_info` in the behaviour
implementing module.

This commit includes several refactorings:
- The waiting acks is now of type `map()` instead of `dict()`
- The ping timer ref is properly managed (before it was part of the
retry mechanism)
- The unacked messages are now stored in a separate map (before they
were stored inside the waiting acks dict).
- Refactorings to simplify the use of the retry mechanism.
- The draining of the offline queue after receiving the CONNACK is now using only function calls instead of message passing.


dergraf:
@monoton This should further improve the bridge. It would be great if you could do some quick tests on this PR. Thanks.

monoton:
what kind of tests do you expects?

dergraf:
similar to the ones you did for the offline queueing.

monoton <notifications@github.com> schrieb am Do., 18. Okt. 2018, 16:01:

monoton:
[behavior]
broker[A] on node[A]
broker[B] on node[B]
bridge from broker[A] to broker[B]
publisher push msgs to broker[A] into topic [test_bridge]
subscriber connected to broker[B] on topic [test_bridge]

[tests]
send msgs betweeen brokers... PASS
stop broker[B] -> send msgs -> rise broker[B] -> receive msgs... `PASS`
drop traffic from broker[A] -> send msgs -> allow traffic from broker[A] -> receive msgs... `PASS`

[little issue]
You renamed `vmq-admin bridges show` -> `vmq-admin bridge show`
but forgot about help
```
  Sub-commands:
    node        Manage this node
    cluster     Manage this node's cluster membership
    session     Retrieve session information
    plugin      Manage plugin system
    listener    Manage listener interfaces
    metrics     Retrieve System Metrics
    api-key     Manage API keys for the HTTP management interface
    trace       Trace various aspects of VerneMQ
    bridges     Manage bridges
    script      Manage lua scripts
  Use --help after a sub-command for more details.
```

dergraf:
@monoton thanks! Fixed the wording. Let's wait for a quick review by @larshesel so this PR will make it into 1.6

dergraf:
@larshesel with this PR the bridge tests are passing deterministic again

dergraf:
@gausby those changes might be interesting for your MQTT client

PR #912, opened 2018-10-25
Add HTTP healthcheck endpoint

codeadict:
Added a `/health` endpoint to the default **http_modules** setting.
This endpoint will allow for better deployments on load-balanced
environments and VerneMQ will need to meet the following conditions:

1. Ensure that all the configured listeners are un running status.
2. Node has joined the cluster and is running.

In case they are both met, the endpoint returns:

```http
GET {{SERVER}}/health HTTP/1.1
Content-Type: application/json
Status-Code: 200
{
  "status": "OK"
}
```

In case the conditions are not met. it can return:

```http
GET {{SERVER}}/health HTTP/1.1
Content-Type: application/json
Status-Code: 503
{
  "status": "DOWN",
  "reason": "Node has not joined the cluster"
}
```

or:

```http
GET {{SERVER}}/health HTTP/1.1
Content-Type: application/json
Status-Code: 503
{
  "status": "DOWN",
  "reason": "Listeners are not ready: mqtt"
}
```

Still a **WIP**, need to add more lager debugs and tests, @dergraf @larshesel feel free to review and recommend something earlier.


larshesel:
Hey - I think this is really cool.

One concern, however, I would imagine it to be important to be able to know if the health of a single node is OK and if the cluster as a whole is OK without having to parse the payload  -- or is there a standard for how payloads should look?

codeadict:
There is not a standard really, the load balancers only know 200 is ok and anything else is bad. What people do parse payloads is usually use curl in combination with https://stedolan.github.io/jq/ for readiness probe. This is what I've seen in AWS ECS and Kubernetes, somebody else may have a different idea and will be nice to hear from it.

larshesel:
So, wouldn't it be an idea to either make it configurable or have multiple paths like `health` and `health-cluster` ?

Personally I would likely want to have the load balancer know about the health of each individual node so it can make intelligent routing decisions. If it only knows about the state of the entire cluster then that wouldn't work.

Maybe @dergraf or @ioolkos have some input here as well?

codeadict:
Yeah, my ideas are that if either the node listeners are or has not joined the cluster successfully, we should not route traffic to that node. Having 2 endpoints will be kind of weird because some load balancers don't allow to have multiple health checks on the same protocol (Looking at AWS :( ).

codeadict:
Any thoughts on this @dergraf @ioolkos?

dergraf:
@codeadict thanks for your efforts

apps/vmq_server/src/vmq_health_http.erl
```diff
@@ -0,0 +1,97 @@
...
+
+-spec cluster_status() -> ok | {error, Reason :: string()}.
+cluster_status() ->
+  ThisNode = node(),
+  try
+    case vmq_cluster:status() of
+      [] ->
+        {error, "Unknown cluster status"};
+      Status ->
+        [{ThisNode, Running}|_] = Status,
```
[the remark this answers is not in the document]
codeadict: Very true, will fix. We only care about the current node, given that each node has its own health-check

changelog.md
```diff
@@ -358,14 +361,14 @@
...
-- Fix bug where an aborted connection handshake caused queued messages not 
+- Fix bug where an aborted connection handshake caused queued messages not
   being migrated properly to remote node.
 
 ## VERNEMQ 1.0.0
 
 - To make the `vmq-admin` tool more consistent, the following changes have been
   made:
-  
+
```
larshesel: Could you revert all the white space changes? I've configured my editor to only remove trailing white space on the line I actually edit, maybe your editor has a similar feature?

codeadict: Yep, i need to fix that, what editor do you use BTW?

larshesel: emacs - for good and for worse ;)

codeadict: For good I guess, it has the best Erlang support, I just cannot unlearn VIM :)

larshesel: VIM has pretty good Erlang support as well I think - I'm sure @dergraf would agree

apps/vmq_server/src/vmq_health_http.erl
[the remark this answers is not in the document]
codeadict: Made these changes @dergraf

codeadict:
Any objection to introducing meck? Some test coverage is hard to get without mocking some function calls, I think it could be of benefit.

larshesel:
> Any objection to introducing meck? Some test coverage is hard to get without mocking some function calls, I think it could be of benefit.

I'm not a fan of meck - my experience with meck is that it's too easy to make brittle tests. That said it depends on how it's used. Mostly I prefer to rewrite code to make it testable if possible. Just my 2 cents.

larshesel:
> I think status codes 200 and 503 are good, we should keep the returned JSON as experimental for now, and see what makes sense in the long run, maybe we can make this adjustable at some point.

@dergraf what about the semantics of the 503 return code - shouldn't it be possible to know if the node is OK or the cluster is OK? I guess we can later introduce some config which makes it configurable if 503 means node or cluster is ok?

larshesel:
> @dergraf what about the semantics of the 503 return code - shouldn't it be possible to know if the node is OK or the cluster is OK? I guess we can later introduce some config which makes it configurable if 503 means node or cluster is ok?

Alright, I hadn't checked that we only return 503 in the cluster case if the node itself is not yet part of the cluster or not reachable - I thought it was any node... 

Sorry for the misunderstanding.

larshesel:
:tada:

codeadict:
Thanks all! ❤️

PR #934, opened 2018-11-01
Fix `max_message_rate` for mqttv5 sessions

larshesel:
The mechanism for calculating the message rate didn't include data
from MQTT 5.0 sessions.


apps/vmq_server/src/vmq_metrics.erl
```diff
@@ -516,10 +516,12 @@ code_change(_OldVsn, State, _Extra) ->
 %%===================================================================
 %%% Internal functions
 %%%===================================================================
-calc_rate_per_conn(REntry, _Entry, 0) ->
+calc_rate_per_conn(REntry, _Entryies, 0) ->
```
dergraf: typo.. Entries

PR #941, opened 2018-11-02
Upgrade node package to fix #644

### LINKED ISSUE #644: vmq-admin trace process doesn't correctly terminate
### Environment

- VerneMQ Version: 1.3.1
- OS: Centos7
- Erlang/OTP version (if building from source): 19.3

### Expected behavior

It should be possible to do a trace, kill the trace (CTRL-C) and start another trace.

### Actual behaviour

When killing a trace invoked using `sudo vmq-admin trace client client-id=<client-id>` the process isn't killed. Example:

```
$ sudo vmq-admin trace client client-id=badauth 
No sessions found for client "badauth"
^C
Session terminated, killing shell... ...killed.
[lhc@node0 ~]$ sudo vmq-admin trace client client-id=badauth 
Cannot start trace as another trace is already running.
```

And the escript process is still running:
```
$ ps -ef | grep trace
vernemq   8641     1  0 06:51 ?        00:00:00 /usr/lib64/vernemq/erts-8.3.5.3/bin/beam -B -fnu -- -root /usr/lib64/vernemq -progname erl -- -home /var/lib/vernemq -- -boot start_clean -noshell -run escript start -extra /usr/lib64/vernemq/erts-8.3.5.3/bin/nodetool -name node0@127.0.0.1 -setcookie vmq rpc_infinity vmq_server_cli command vmq-admin trace client client-id=badauth
```

The workaround is to kill this process and another trace can be started.

This doesn't happen with developer builds.

larshesel:
see vernemq/node_package#4 for details.

fixes #644


rebar.config
```diff
@@ -16,7 +16,7 @@
         {cuttlefish, {git,"git://github.com/kyorai/cuttlefish.git", {branch, "develop"}}},
         {vernemq_dev, {git, "git://github.com/erlio/vernemq_dev.git", {branch, "master"}}},
```
dergraf: maybe we should change the repo for vernemq_dev as well

larshesel: Yeah, I'll do this in another PR

PR #943, opened 2018-11-06
Fix blocking cross node publish

dergraf:


larshesel:
Did a lengthy 3-node cluster test and everything checks out: no publishers are blocked and taking a node down doesn't increase the load on the other nodes

apps/vmq_server/src/vmq_cluster_node.erl
```diff
@@ -174,21 +178,25 @@ handle_message({msg, CallerPid, Ref, Msg}, State) ->
...
-    NewTRef =
-    case TRef of
-        undefined ->
-            reconnect_timer();
-        _ ->
-            %% we're already reconnecting
-            TRef
-    end,
-    State#state{reachable=false, reconnect_tref=NewTRef};
+handle_message({connect_async, AsyncPid, {ok, {Transport, Socket}}},
```
larshesel: I think you should rename `connect_async` to something like `connect_finished`. When I read the code the first time I thought `connect_async` was the message sent to start the connect process.

PR #946, opened 2018-11-07
Fix duplicate Qos2 publish issue (#944)

### LINKED ISSUE #944: QOS 2 messages duplicated
### Environment

- VerneMQ Version: 1.6.1
- OS: CentOS 7

```
allow_anonymous = on
allow_register_during_netsplit = on
allow_publish_during_netsplit = on
allow_subscribe_during_netsplit = on
retry_interval = 20
max_online_messages = 100000
max_offline_messages = 10000
log.console.level = debug
erlang.distribution.port_range.minimum = 6001
erlang.distribution.port_range.maximum = 7999

leveldb.write_buffer_size_min = 120MB
leveldb.write_buffer_size_max = 160MB
```

### Expected behavior

- VerneMQ should not deliver a QOS 2 message twice to QOS 2 subscribers even if the original sender re-transmitted it with the DUP flag.

#### Summary

We are running [mqtt-malaria](https://github.com/remakeelectric/mqtt-malaria) against a VerneMQ cluster. Under bigger load (10 concurrent publishers, 10k messages each), we get duplicate messages on the subscriber side (there is only one subscriber in the test).

#### Details

We are running a cluster of 4 VerneMQ nodes with IP addresses: 192.168.122.{166,40,236,250}. The host which runs malaria and the subscriber has IP 192.168.122.1.

Here is how we launch the subscriber process:

```mosquitto_sub  -v -h 192.168.122.166 -i sk-3 -t '#' -q 2 > /tmp/w```

Here is how we run malaria:

```./malaria publish -H 192.168.122.40 -n 10000 -P 10 -q 2 -t```

When malaria finishes and all messages are delivered to the subscriber, we count the total number of messages in the file and compare the result with the number of unique messages received:

```
$ wc -l /tmp/w
131292 /tmp/w
$ sort  /tmp/w | uniq | wc -l
100000
```

We [recorded the network traffic](https://github.com/vernemq/vernemq/files/2553641/malaria-10x10k-3-export-2.txt) using wireshark and traced one particular message which was seen twice by the subscriber. It is the first message in the dump file (msgid: 3022, topic: mqtt-malaria/beem...-0/data/3022/10000). Here is a short outline to ease getting through the dump file:

1. Client sends PUBLISH (l. 8)
2. Now, possibly concurrently, two things happen:
2.1. Server responds with PUBREC (l. 75)
2.2. Client sends PUBLISH of the same message with the same message ID but with DUP flag set (l. 93)
3. Client sends PUBREL (l. 162)
4. Server sends PUBREC (l. 260)
5. Server sends PUBCOMP (l. 1936)
6. Client sends PUBREL (l. 2002)
7. Server sends PUBCOMP (l. 3624)
8. Server sends PUBLISH message to the subscriber (l. 4466)
9. Server sends PUBLISH message to the subscriber once again (l. 4553, different message ID, no DUP flag set).

We think that since the server received the second PUBLISH (with DUP flag) before PUBREL, it should be able to check that this particular msg ID is already being handled and continue normal procedure. It shouldn't start a new QOS 2 flow with the client. Would you agree?

larshesel:
Fixes #944

### DIFF
--- a/apps/vmq_server/src/vmq_mqtt5_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt5_fsm.erl
@@ -1269,31 +1269,37 @@ dispatch_publish_qos2(_, _, error, _) ->
 dispatch_publish_qos2(MessageId, Msg, Cnt, State) ->
     #state{username=User, subscriber_id=SubscriberId,
            cap_settings=CAPSettings, reg_view=RegView, waiting_acks=WAcks} = State,
-
-    case publish(CAPSettings, RegView, User, SubscriberId, Msg, State) of
-        {ok, NewMsg, SessCtrl, NewState} ->
+    case maps:get({qos2, MessageId}, WAcks, not_found) of
+        not_found ->
+            case publish(CAPSettings, RegView, User, SubscriberId, Msg, State) of
+                {ok, NewMsg, SessCtrl, NewState} ->
+                    Frame = #mqtt5_pubrec{message_id=MessageId, reason_code=?M5_SUCCESS, properties=#{}},
+                    _ = vmq_metrics:incr({?MQTT5_PUBREC_SENT, ?SUCCESS}),
+                    {NewState#state{
+                       fc_receive_cnt=Cnt,
+                       waiting_acks=maps:put({qos2, MessageId}, {Frame, NewMsg}, WAcks)},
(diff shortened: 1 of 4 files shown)


larshesel:
Merged this without waiting for travis - it had been stuck all night and didn't seem to want to restart this morning either...

PR #949, opened 2018-11-09
Log peer information on authentication failures

### LINKED ISSUE #931: Adapt Logs to be useful for fail2ban
I guess at least for failed authentications.

larshesel:
Fixes #931

### DIFF
--- a/apps/vmq_server/src/vmq_mqtt5_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt5_fsm.erl
@@ -860,8 +860,8 @@ check_user(#mqtt5_connect{username=User, password=Password, properties=PropsIn0}
                     register_subscriber(F, OutProps, QueueOpts,
                                         NewState#state{session_expiry_interval=SessionExpiryInterval});
                 {error, no_matching_hook_found} ->
-                    lager:error("can't authenticate client ~p due to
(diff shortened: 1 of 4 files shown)


larshesel:
Fixed dialyzer issue (thanks, dialyzer, that was a real bug!)

larshesel:
> Shouldn't we use the IP:Port format in the logs?

That was my initial thought, but how would you format the IPv6 strings with ports. It seems that in URLS you'd do `[2a02:120b:7fd:1060:9eb6:d0ff:fed0:232f]:12744`, but I'm not sure if it is really a standard. That's why I went with a space between the IP and port. If you think it better I'll change it to `IPv4:Port` and `[IPv6]:Port`.

larshesel:
> If you think it better I'll change it to `IPv4:Port` and `[IPv6]:Port`.

You know what, I'll just do that... The other format annoyed me as well.

dergraf:
good point.let's be pragmatic, what is the format that yields the easiest regex pattern. ;)

larshesel:
> good point.let's be pragmatic, what is the format that yields the easiest regex pattern. ;)

I don't know, you tell me! :)

larshesel:
Changed it to `IPv4:Port` and `[IPv6]:Port`.

Issue #926, opened 2018-10-30
Ranch listener crash when clinet connected with some offline message


boboxiaodd:
### Environment

- VerneMQ Version: 1.6.0
- OS: centos7
- Erlang/OTP version (if building from source):

include Version 1.4.0 , 1.5.0
### Expected behavior
```erlang
vmq_mqtt_fsm:connected:296 Ranch listener {{0,0,0,0},9000} terminated with reason: no case clause matching {mqtt_pubrel,12} in vmq_mqtt_fsm:connected/2 line 296
```

dergraf:
Thank you @boboxiaodd for opening this issue.
I am a bit puzzled by this, as I expected it to be fixed by https://github.com/erlio/vernemq/issues/926.
However, something might be missing..

Is it possible that this error is triggered by 'persisted' data, that was persisted back when you were running 1.4.0 ? Could you reproduce the error if you start with an 'empty' state?

boboxiaodd:
@dergraf 
 i use   1.4.1  found the issues , and update to 1.6.0 , still  found the issues

boboxiaodd:
@dergraf  by the way:
In version: 1.6.0

i found someone use http protocol to visit vernemq ,and Ranch listener terminated:
```erlang
vmq_mqtt_pre_init:determine_protocol_version:83 Ranch listener {{0,0,0,0},9000} terminated with reason: no function clause matching vmq_mqtt_pre_init:determine_protocol_version(<<"GET / ... ">>)
```
i found the reason:

https://github.com/erlio/vernemq/blob/445fa6aa6d5b69d7206c5c8aed1d45f257ec0dfd/apps/vmq_server/src/vmq_mqtt_pre_init.erl#L83

boboxiaodd:
@dergraf 
i check another vernemq server version 1.5.0
i don't have Retain message
all of topics are qos2
some of error:
```erlang
2018-10-30 17:23:45.848 [error] <0.25802.25> CRASH REPORT Process <0.25802.25> with 0 neighbours crashed with reason: no case clause matching {mqtt_pubrel,1} in vmq_mqtt_fsm:connected/2 line 310
2018-10-30 17:23:45.848 [error] <0.25802.25> Ranch listener {{0,0,0,0},1883} terminated with reason: {case_clause,{mqtt_pubrel,1}}
[...]

PR #956, opened 2018-11-14
ignore duplicate pubrec

boboxiaodd:
https://github.com/vernemq/vernemq/issues/926

### DIFF
--- a/apps/vmq_server/src/vmq_mqtt_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt_fsm.erl
@@ -316,7 +316,7 @@ connected(#mqtt_pubrec{message_id=MessageId}, State) ->
                waiting_acks=maps:update(MessageId, PubRelFrame, WAcks)},
             [PubRelFrame]};
         #mqtt_pubrel{} ->
-            lager:debug("Duplicate pubrel ~p ,ignore", [MessageId]),
+            lager:debug("Duplicate pubrec ~p ,ignore", [MessageId]),
             {State,[]};
         not_found ->
             lager:debug("stopped connected session, due to qos2 puback missing ~p", [MessageId]),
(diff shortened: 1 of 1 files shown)


apps/vmq_server/src/vmq_mqtt_fsm.erl
```diff
@@ -315,6 +315,9 @@ connected(#mqtt_pubrec{message_id=MessageId}, State) ->
                retry_queue=set_retry(pubrel, MessageId, RetryInterval, RetryQueue),
                waiting_acks=maps:update(MessageId, PubRelFrame, WAcks)},
             [PubRelFrame]};
+        #mqtt_pubrel{} ->
+            lager:debug("Duplicate pubrec ~p ,ignore", [MessageId]),
+            {State,[]};
```
larshesel: I've been looking into thee issue a bit further and think I understand it - it does seem we don't handle it correctly when the client resends a pubrec, but we can't ignore it, the spec says this (ref [MQTT-4.3.3-1] in docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html):

```
MUST send a PUBREL packet when it receives a PUBREC packet from the receiver. This PUBREL packet MUST contain the same Packet Identifier as the original PUBLISH packet.
```

So I think you're on the right track but I think the correct behaviour would be to do something like:

```erlang
        #mqtt_pubrel{} = Frame ->
            {State,[Frame]};
```

Which would cause the pubrel to be re-sent. I'll write up a test reproducing this on MQTT 3.1.1 and MQTT 5.0 so we can never break this again.

larshesel: We probably also need to fix this in the MQTT 5.0 fsm

larshesel: /cc @dergraf - just FYI

larshesel:
Replaced by #957

PR #963, opened 2018-11-18
Fix updating properties on publish

manifest:
At the moment it is impossible to update **user properties** from a **publish hook**.


larshesel:
Hi @manifest - you're absolutely right - so far we'd only tested passing the properties through when publishing, not modifying them.

I added a small test which I plan to extend for other properties we want to be able to modify and also added a small changelog item for it. I pushed it to your branch. 

Is this fine for you, then I'd squash and merge!

manifest:
Hi @larshesel! It's perfectly fine for me :-) I appreciate your help.

PR #965, opened 2018-11-20
Update THANKS file

larshesel:
Hi @manifest 

I'd like to add you to the THANKS file - ok for you? :)


manifest:
Sure. It would be an honor :-)

larshesel:
hehe, we're super happy for the feedback and the contributions! and please keep it coming - the MQTTv5 support is still new and some things are still not exactly where they should be, but we're working on the polish, but really feedback from actual users is just the best! So thanks!

Issue #974, opened 2018-11-22
auth_on_publish / auth_on_subscribe - disconnect client (should it? could it?)


xbugster:
### Environment

- VerneMQ Version: 1.5.0-127-ge130553
- OS: ubuntu linux 16.04
- Erlang/OTP version (if building from source): erts-8.3.5.5@OTP 19

### Expected behavior

Hello, dear friends.
We are again dealing with LUA plugin, this time we are trying to tinker with `auth_on_publish` and `auth_on_subscribe`, we expect the ability to disconnect client who is taking illegal action (illegal topic in our case).
we made sure, that we are 100% returning false on both hooks.

### Actual behaviour

Using mosquitto clients, subscriber does not gets disconnected when false returned from `auth_on_subscribe`, and it seems the same happens with `auth_on_publish`.
SUB client hangs on
```
Client mosqsub|13851-bountyhun sending CONNECT
Client mosqsub|13851-bountyhun received CONNACK (0)
Client mosqsub|13851-bountyhun sending SUBSCRIBE (Mid: 1, Topic: ing/response/xbugster/prj/someid, QoS: 0)
Client mosqsub|13851-bountyhun received SUBACK
Subscribed (mid: 1): 0
```

and in case of publish to illegal topic(which means FALSE returned by auth) the client still on connect.
**both of clients does not send, nor receive messages, but the connection is open**

Our question, whether we can disconnect client using on of those hooks ? We have seen undocumented feature existing for erlang plugin users `self() ! disconnect` or similar.

Thank you for your help.
my teammate will add more information from using other client responses/behavior.

larshesel:
A client will *never* get disconnected if it tries to subscribe to an illegal subscription. As per the spec the broker simply informs the client in the SUBACK that the subscribe was unsuccessful.

An MQTT 3.1 client will stay connected when `auth_on_publish` returns false, while and MQTT 3.1.1 client will  be disconnected. From the above I expect you're using an MQTT 3.1 (protocol level 3) client. You can check using the tracing command I mentioned in an earlier issue.

Disconnecting a client is now part of our developer API: https://github.com/vernemq/vernemq_dev/blob/master/src/vernemq_dev_api.erl#L27, but this isn't exposed to Lua scripts at the moment.

xbugster:
Cool. You guys, are the best, as always!
Thank you for great explanation! 
Looking forward for client disconnection api.

PR #977, opened 2018-11-26
Expose disconnect api to lua

larshesel:
Expose disconnect client api in lua as mentioned in #974 

Biggest annoyance here is that we can't well test the functionality as we don't have access to a running `vmq_server` from `vmq_diversity` tests. I ad-hoc tested this and added a `syntax` test.

Note I discovered a bug in the dev api: https://github.com/vernemq/vernemq_dev/pull/9


larshesel:
ping @dergraf , what do you think about this?

PR #981, opened 2018-11-27
Prepare 1.6.2

larshesel:
Would like confirmation of a bugfix (https://github.com/vernemq/vernemq/pull/971) before finishing up


larshesel:
> Would like confirmation of a bugfix (#971) before finishing up

Never got the confirmation - I think we should just cut 1.6.2 and include any refinements of the above in a possible next bugfix-release.

@dergraf let me know if this is OK then I'd do the normal release preparations and tagging and merge this to the 1.6-bugfix branch.

PR #985, opened 2018-11-29
Improve `vmq_webhooks` errors

larshesel:
This will mean we will for example log:

2018-11-29 10:16:55.314 [error] <0.468.0>@vmq_mqtt_fsm:auth_on_publish:712 can't auth publish [undefined,{[],<<"client">>},1,[<<"my">>,<<"topic">>],<<"hello">>,false] due to <<"not allowed">>

where the reason is <<"not_allowed">> instead of just `error.


dergraf:
what is the idea behind using a binary instead of an atom for the error reason?

larshesel:
> what is the idea behind using a binary instead of an atom for the error reason?

The idea is that this comes from the endpoint and as such converting it to an atom can be a bad idea.

larshesel:
Of course that could be maybe be improved to converting to the atoms we know, but to me this seems a lot better than just returning `error`

PR #990, opened 2018-12-03
SWC third iteration, brings many bugfixes and improvements

dergraf:
This PR adds SWC as a first class plugin to VerneMQ and can be enabled using `metadata_plugin = vmq_swc` inside `vernemq.conf`.

By default it uses the LevelDB database, as this is the default backend used by the message store, the dependency is likely already fulfilled. `vmq_swc` is already prepared to support other backends like RocksDB and Leveled (a Erlang based LevelDB variant). 

Other important changes:
- remove the special FSM for the Full Sync case and use the 'normal' anti-entropy exchange FSM
- the change above required to adapt the Dotkeymap to also keep the actvie set of dots.
- the dotkeymap implementation received a complete rewrite, including a triq quickcheck model to ensure we're not losing any dots during incremental garbage collection.


dergraf:
@ioolkos checkout this branch, run `make rel` and set the `metadata_plugin = vmq_swc` inside `vernemq.conf` that's all ;)

dergraf:
I am quite sure the Erlang 20.3 on Travis CI was using the buggy gen_statem resulting in reproducible SWC test fails... using the 20.3.8 let the tests pass, finally.

dergraf:
@larshesel let's start a peer review session soon. You want me to squash it now or afterwards?

larshesel:
yes, let's! And let's squash afterwards.

PR #994, opened 2018-12-05
Add histogram metric type

dergraf:


apps/vmq_server/src/vmq_server_cli.erl
```diff
@@ -143,32 +143,53 @@ vmq_server_metrics_cmd() ->
...
+                                        end, HistogramLines, Buckets)
+
                               end,
-                          Line = [SType, ".", SMetric, " = ", SVal],
                           case Describe of
                               true ->
-                                  [clique_status:text(lists:flatten(["# ", Description])),
-                                   clique_status:text(lists:flatten([Line, "\n"]))|Acc];
+                                  [clique_status:text(lists:flatten(["# ", Description]))
+                                   | [[[Lines | clique_status:text("\n")]] | Acc]];
```
larshesel: Fails on dialyzer here. I created a PR to this branch https://github.com/dergraf/vernemq/pull/4 fixing that and there tried to clean up a bit so we only return iolists in all branches and `clique_status:text/1` and `lists:flatten/1` then only have to be called once.

Note, I haven't tested the output of the histogram data in the cli as I didn't know how to easily get such a metric.

dergraf: The output will look similar to this:

```
histogram.metadata_get_microseconds_bucket_infinity = 176
histogram.metadata_get_microseconds_bucket_1000000 = 176
histogram.metadata_get_microseconds_bucket_100000 = 176
histogram.metadata_get_microseconds_bucket_10000 = 176
histogram.metadata_get_microseconds_bucket_1000 = 175
histogram.metadata_get_microseconds_bucket_100 = 172
histogram.metadata_get_microseconds_bucket_10 = 0
histogram.metadata_get_microseconds_count = 176
histogram.metadata_get_microseconds_sum = 7279
```

dergraf: Tried it already with the swc-leveled branch

dergraf: The implemented histogram is a cumulative histogram, as this simplifies the handling for Prometheus (I guess, because they expect the histogram to be cumulative)

PR #995, opened 2018-12-07
Expose bridge metrics

larshesel:
So far we only have a metric for dropped messages due to the queue
being full.

### DIFF
--- a/apps/vmq_commons/src/gen_mqtt_client.erl
+++ b/apps/vmq_commons/src/gen_mqtt_client.erl
@@ -29,6 +29,8 @@
            {gen_fsm,cancel_timer,1}]}]).
 -endif.
 
+-define(PD_QDROP, pd_queue_drop).
+
 -callback init(Args :: any()) ->
     {ok, State :: any()} |
     {ok, State :: any(), Extra :: any()} |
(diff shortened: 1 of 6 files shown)


apps/vmq_bridge/priv/vmq_bridge.schema
```diff
@@ -265,6 +266,21 @@
          Term
  end}.
 
+{mapping, "vmq_bridge.vmq_metrics_mfa", "vmq_bridge.vmq_metrics_mfa",
```
dergraf: not sure if we really need this to be in the schema...thoughts?

apps/vmq_commons/src/gen_mqtt_client.erl
```diff
@@ -161,6 +164,10 @@ start(Name, Module, Args, Opts) ->
 info(Pid) ->
     gen_fsm:sync_send_all_state_event(Pid, info).
 
+stats(Pid) ->
+    [{dictionary, PD}] = erlang:process_info(Pid, [dictionary]),
```
dergraf: :1st_place_medal:

larshesel: You like that, huh?

apps/vmq_bridge/priv/vmq_bridge.schema
larshesel: adding it the app.src file only isn't enough(I thought it was?) - it's anyway hidden and won't be in the config file.

dergraf: to my knowledge app.src is sufficient... I tested it with my SWC release.

larshesel: You're right - I'd done something wrong when testing. Removed it.

changelog.md
```diff
@@ -46,7 +46,7 @@
   reusing already available data.
 - Enable plugins to expose metrics via the available metric providers.
 - Add histogram metric type.
-- Fix incorrect debug log message in QoS 2 flow.
```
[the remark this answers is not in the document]
dergraf: Otherwise good to be merged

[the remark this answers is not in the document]
larshesel: oops, think I removed the wrong line there when rebasing... good catch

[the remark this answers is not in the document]
larshesel: fixed.

Issue #988, opened 2018-11-30
High memory usage even after test finishes


szymonk:
### Environment

- VerneMQ Version: 7ea2de2bd4db01cbe9e678ef2653a831d7e3cf9a
- OS: CentOS 7
- Erlang/OTP version (if building from source): 21.1
- VerneMQ configuration (vernemq.conf) or the changes from the default

```
allow_anonymous = off
allow_register_during_netsplit = on
allow_publish_during_netsplit = on
allow_subscribe_during_netsplit = on
allow_unsubscribe_during_netsplit = off
allow_multiple_sessions = off
max_inflight_messages = 100
max_online_messages = 10000
max_offline_messages = 1000
max_message_size = 0
upgrade_outgoing_qos = off
listener.max_connections = infinity
listener.nr_of_acceptors = 10
systree_enabled = off
erlang.async_threads = 64
erlang.max_ports = 262144
leveldb.maximum_memory.percent = 8
erlang.distribution.port_range.minimum = 6000
erlang.distribution.port_range.maximum = 7999
erlang.distribution_buffer_size = 10MB
plumtree.outstanding_limit = 50000
```

We also set +hmqd off_heap on erlang VM.

### Expected behavior

VerneMQ should need amount of memory which could be described in terms of parameters such as numer of clients, subscriptions, messages, nodes, etc.

### Actual behaviour

Usually memory usage on our VerneMQ nodes is about 6 GB. However, sometimes we are observing excessive usage of RAM that gets worse over time. It happens that 16 GB RAM + 32 GB of swap is not enough for handling 80k connections and vernemq is killed by the OS. 

#### Details

Our setup:

- 5 VerneMQ instances, 16GB RAM, 4 CPU each
- haproxy as a load balancer (on a separate machine)
- 28 mzbench nodes serving as MQTT clients

Test scenario:

- 400k clients (let's call them "consumers") connect to the cluster over 9 minutes with a constant rate. Each of them subscribes to a topic which is unique per client.
- 1 special client ("producer") connects to the cluster. Once all the consumers are connected it starts to send messages to consumers' topics, one message per topic (client), 800 messages per second. The producer also subscribes to its own topic. This step of the test lasts for 15 minutes.
  - Once a consumer receives a message from the producer, it sends the message back to the producer.
- When the exchange of the messages is finished, all the clients gradually disconnect.

Here are grafana charts showing memory and swap availability as well as network traffic and number of open tcp connections.
[...]

PR #1009, opened 2018-12-19
Protect from subscription flood (vernemq/vernemq#988)

szymonk:
Switch to plumtree version which supports drop_i_have_threshold. This
allows to drop some messages instead of accumulating them infinitely
and not having CPU time to process them.

Since the modified plumtree version needs newer gen_server2, update it
globally.


larshesel:
This looks good to me - I think it needs a changelog at the bottom of the current unreleased fixes (above 1.6.0). Maybe sth like:

```
- Add protection mechanism for the plumtree metadata store from subscription floods by dropping and not acknowledging `i_have` messages when the mailbox size grows above a certain threshold. This threshold is configurable via the `plumtree.drop_i_have_threshold` hidden option. The default is 1000000.
```

/cc @dergraf

larshesel:
Thank you so much for this contribution!

larshesel:
I completely forgot, may I add you to the THANKS.md file? Do you have other names than @szymonk  and @p3we - let me know what you prefer! Cheers!

szymonk:
That would be great! We would prefer to be added as Artur Sobierak and Szymon Kałasz. Also, to credit the company we work for, could you please add a mention to `changelog.md` that this improvement has been contributed by ADB (https://www.adbglobal.com)?

larshesel:
absolutely! will prepare a PR right away

larshesel:
Something like this, would that be fine? https://github.com/vernemq/vernemq/pull/1014

larshesel:
btw, we've been thinking a bit about how to improve the metadata throughput and we have some preliminary work in #1012 but there are still some challenges to overcome there. Maybe this would also be of interest to you when it's finished. The idea / observation is that usually when a client connects it also immediately subscribes. Currently this results in two writes to the metadata store (one that the client is there and the second that it now has a subscription), but if these two writes can be combined into one we can save 50% of the writes to the metadata store...

szymonk:
Looks good!
The connect/subscription optimizations sounds interesting. Thanks for the info!

PR #1012, opened 2018-12-20
WIP: Coalesce metadata

larshesel:
this is a WIP which is able to merge writes for the same subscriber to the metadata store when the logical writes happen temporally near each other.

The big outstanding issue is that this breaks a lot of tests since subscriptions, which are already handled in an eventually consistent manner, now are written slightly later which is just enough to make the tests break.

So there are two approaches: 
- fix the tests (most of them publish to a topic which hasn't been inserted into the local routing table yet) by waiting for the relevant data to have been propagated.
- insert the subscriptions directly into the local routing table which would mean locally you will start receiving publishes matching the subscriptions as soon as the SUBACK is returned.

The latter seems to me the best approach, but it's not clear how to best implement this. Currently the routing table `vmq_reg_trie` is manipulated by sending it a set of old subscriptions and a set of new subscriptions and then calculating what is to be removed and what is to be added. The optimizer currently only has access to the latest value, not the old one. Two options presents itself: the optimizer can keep track of added/removed topics (SUB/UNSUB) somehow or keep track of the old (where to get this without extra cost?) and new value....

### DIFF
--- a/apps/vmq_server/src/vmq_coalesce.erl
+++ b/apps/vmq_server/src/vmq_coalesce.erl
@@ -16,7 +16,8 @@
 -behaviour(gen_server).
 
 %% API
--export([start_link/0]).
+-export([start_link/0,
+         start_link/1]).
 
 -export([put/3,
          get/2,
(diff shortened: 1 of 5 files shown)


larshesel:
Closing this for now

PR #1021, opened 2019-01-02
[WIP]: Upgrade to Cowboy 2.5.0 and Ranch 1.6.2

codeadict:
**Work In Progress:** Ranch has been updated and MQTT listener and the VMQ
Cluster communication are working fine, most handlers for HTTP listeners are also
working. Still need to fix:

- [x] API Handler.
- [ ] Websockets Listener.

Need to run automated tests and also do extensive testing for multiple
scenarios manually or using some scripts because there are many things
that have changed.

*Note:* These are not the latest versions of these packages but the ones
before the native proxy protocol was added, once this is done we will
upgrade in a separate PR and remove the proxy protocol implementation
from `vmq_ranch_proxy_protocol.erl`.


dergraf:
Note to myself: Initial review looks good... but needs some manual testing.

codeadict:
Cool, mostly needs the fix for the WebSockets interface, the `cowboy_req:get` api is not available anymore, so is hard to get the raw TCP socket, will figure out how to do this one

larshesel:
I found this regarding `cowboy_req:get`: https://github.com/ninenines/cowboy/issues/1147 and indeed, the `sock/1` and `cert/1` seem to exist: https://github.com/ninenines/cowboy/blob/master/src/cowboy_req.erl#L181-L188

codeadict:
ya I fixed one `cowboy_req:get` with `cowboy_req:cert` and indeed `sock/1` gives you the ip and port, so may need to change that functionality a bit, another option is when starting the listener add the socket from ranch to the request env

codeadict:
I'll play with this part tonight, will not be that hard

larshesel:
Hi @codeadict I'm looking into finishing this up - as far as I can see the outstanding work is in the websoket code. As far as I can tell the biggest outstanding issue is getting the websocket code going. Do you remember if there were any other things on the list. I'll try and see if I can fix the missing `Type` in `vmq_websocket`.

codeadict:
You are right, `vmq_websocket` is the main thing that is not working, the Proxy protocol needs also to be tested well.

larshesel:
I've rebased your branch and pushed it to my fork: https://github.com/larshesel/vernemq/tree/cowboy_25

I've upgraded to cowboy 2.6.3 and implemented support for the proxy protocol - only real outstanding issue is to fetch the socket so we can set socket options in the websocket handler - and of course testing the whole thing :)

larshesel:
The proxy stuff was pretty easy - the harder part is the socket. I implemented a wrapper module for `cowboy_websocket` so I would be able to put the socket into the handler data in the `takeover` function. But unfortunately the `takeover` function isn't called in the wrapper (but in `cowboy_websocket`), so I discussed it with Loïc and he suggested I'd have to implement a stream handler. This looks more complicated, at least on the surface.

larshesel:
Closing this in favor of #1205

PR #1035, opened 2019-01-16
WIP: Make Lua states restartable

larshesel:


larshesel:
hey @dergraf the tests fail because the state manager hasn't always been registered with the plugin before the tests start, I did this ugly hack for that: https://github.com/vernemq/vernemq/pull/1035/commits/e415a92d365612745d3369e633c297ca1c7da0ee - got a better idea?

PR #1037, opened 2019-01-17
Fix timing issue in shared subscriber tests

larshesel:
Working on sth else I got tired of the shared sub tests crapping out from time to time. The problem was that we didn't wait until the client was actually offline before starting to publish. Hopefully waiting for the offline hook to be invoked should be sufficient.


larshesel:
@dergraf could you take a look at this at some point?

PR #1038, opened 2019-01-17
Increment MQTT4_CONNACK_SENT metric

szymonk:
Hello VerneMQ Team,

VerneMQ seems not to be reporting CONNACK SENT metric correctly for MQTT < 5. This patch fixes it for me. Do you think it is good enough and can be merged?


larshesel:
Hi @szymonk this is a great catch! I added a log line on top of your commit - if the tests passes I'll squash and merge!

I wonder how we missed that - I see we even added those metrics for the MQTT 5.0 fsm...

larshesel:
Thanks a lot for this! :+1:

szymonk:
Thanks for quick merge!

PR #1040, opened 2019-01-18
Early return/event from batch

dergraf:


larshesel:
needs changelog line

larshesel:
I guess the existing one already covers this... forget about the changelog

PR #1046, opened 2019-01-23
Add a new metric to track queue initializations

szymonk:
This PR adds a new queue_init_from_storage counter which is incremented whenever a queue is initialized with messages from the message store.

The use case for this is to monitor progress of queue initialization. Currently there is a queue_setup metric which is helpful but it only tracks the number of queue processes. If the message store is big enough, feeding queues with messages can take significant time. We ran into cases where we exposed the broker too early (when the queues were still being initialized) and as a result the broker's performance was degraded.

The "Queue Init" series in the graph below are derived from the new metric. They may be used to catch the moment when the queues are initialized. In this case it happened 6-7 minutes after the queue processes had been started.

vernemq-queue-init-from-storage-1

### DIFF
--- a/apps/vmq_server/src/vmq_metrics.erl
+++ b/apps/vmq_server/src/vmq_metrics.erl
@@ -61,6 +61,7 @@
          incr_mqtt_error_unsubscribe/0,
 
          incr_queue_setup/0,
+         incr_queue_init_from_storage/0,
          incr_queue_teardown/0,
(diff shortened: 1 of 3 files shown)


szymonk:
Yes, that sounds better. I'll change it.

larshesel:
Thanks! I'll just wait for travis to do its thing then I'll squash and merge!

PR #1047, opened 2019-01-23
Fix queue start timing issue

larshesel:
- switch to only starting new queues when clients connect or when queues are (forcefully) migrated or dead queues fixed.

### DIFF
--- a/apps/vmq_server/src/vmq_reg.erl
+++ b/apps/vmq_server/src/vmq_reg.erl
@@ -22,6 +22,7 @@
          unsubscribe/3,
          register_subscriber/4,
          register_subscriber/5, %% used during testing
+         replace_dead_queue/3,
          delete_subscriptions/1,
          %% used in mqtt fsm handling
          publish/4,
@@ -478,8 +479,16 @@ migrate_offline_queue(SubscriberId, QPid, {[Target|Targets], AccQs, AccMsgs} = A
 
 fix_dead_queues(_, []) -> exit(no_target_available);
 fix_dead_queues(DeadNodes, AccTargets) ->
-    %% DeadNodes must be a list of offline VerneMQ nodes
-    %% Targets must be a list of online VerneMQ nodes
+    %% The goal here is to:
+    %%
+    %% 1. create queues on another node for all clean_start = false
+    %% sessions.
+    %%
+    %% 2. Purge the old nodename from the rest of the metadata to
+    %% ensure publishes are not forwarded to the dead node.
+
+    %% DeadNodes must be a list of offline VerneMQ nodes. Targets must
+    %% be a list of online VerneMQ nodes
     {_, _, N} = fold_subscribers(fun fix_dead_queue/3, {DeadNodes, AccTargets, 0}),
     lager:info("dead queues summary: ~p queues fixed", [N]).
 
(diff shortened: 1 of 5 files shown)


apps/vmq_server/src/vmq_reg.erl
```diff
@@ -492,32 +501,90 @@ fix_dead_queue(SubscriberId, Subs, {DeadNodes, [Target|Targets], N}) ->
...
+        _ ->
+            %% Assume all the subscriptions come with the same clean_session flag.
+            [{_, StartClean, _}|_] = Subs,
+            RepairQueueFun =
+                fun() ->
+                        case rpc:call(Target, vmq_reg, replace_dead_queue, [SubscriberId, DeadNodes, StartClean, Subs]) of
+                            ok ->
+                                {DeadNodes, Targets ++ [Target], N + 1};
+                            Error ->
+                                lager:format("repairing dead queue for ~p on ~p failed due to ~p~n", [SubscriberId, Target, Error]),
```
dergraf: Travis CI is complaining about lager:format, lager:info?

apps/vmq_server/src/vmq_reg.erl
```diff
@@ -492,32 +501,90 @@ fix_dead_queue(SubscriberId, Subs, {DeadNodes, [Target|Targets], N}) ->
...
+            ok;
+        QPid ->
+            %% we force a disconnect to ensure the client reconnects
+            %% and reestablishes consistent metadata.
+            vmq_queue:force_disconnect(QPid, ?ADMINISTRATIVE_ACTION, true),
+            ok
+    end;
+replace_dead_queue(SubscriberId, DeadNodes, _StartClean = false, Subs) ->
+    %% At this point we may have a new subscriber on either this or
+    %% orther nodes and their subscribtions may differ from the one we
```
dergraf: typo `orther`

apps/vmq_server/src/vmq_reg.erl
```diff
@@ -492,32 +501,90 @@ fix_dead_queue(SubscriberId, Subs, {DeadNodes, [Target|Targets], N}) ->
...
+    %% other nodes which may have another set of subscriptions than
+    %% what the old node had.
+
+    %% To ensure we end up with a client which has a consistent state,
+    %% we write an empty subscription to the store which will cause
+    %% clients on other nodes to be disconnected (if
+    %% allow_multiple_sessions=false) (TODO: formalize this behaviour
+    %% in a test-case) and they'll then reconnect.
+    Subs = vmq_subscriber:new(true),
+    vmq_subscriber_db:store(SubscriberId, Subs),
```
dergraf: aren't we create another race with the `get_queue_pid` below?

apps/vmq_server/src/vmq_reg.erl
```diff
@@ -492,32 +501,90 @@ fix_dead_queue(SubscriberId, Subs, {DeadNodes, [Target|Targets], N}) ->
...
+
+    %% check if there's any work to do:
+    NewSubs = rewrite_dead_nodes(Subs, DeadNodes, '$notanodename', false),
+    case Subs of
+        NewSubs ->
+            %% no subs were changed, nothing to do here.
+            {DeadNodes, [Target|Targets], N};
+        _ ->
+            %% Assume all the subscriptions come with the same clean_session flag.
+            [{_, StartClean, _}|_] = Subs,
```
dergraf: shoudn't this be `NewSubs`?

apps/vmq_server/src/vmq_reg.erl
```diff
@@ -492,32 +501,90 @@ fix_dead_queue(SubscriberId, Subs, {DeadNodes, [Target|Targets], N}) ->
...
+                    %% an error and let the caller retry.
+                    error;
+                LocalSubs ->
+                    NewSubs = rewrite_dead_nodes(LocalSubs, DeadNodes, node(), false),
+                    %% store the updated subs, also to make sure to
+                    %% propagate the new values to all other nodes.
+                    vmq_subscriber_db:store(SubscriberId, NewSubs)
+            end;
+        _ ->
+            %% new queue, just store the updated subscriptions without
```
dergraf: `new` or `existing` queue..?

larshesel: New queue, as above the `{ok, true, _OldQPid}` was handled.

apps/vmq_server/src/vmq_reg.erl
larshesel: I think you're right, it would probably be better to put the `store(SId, Subs)`  into the each of the two branches of the case statement.

larshesel:
I was thinking instead of passing in target nodes from `vmq_server_cli` it would be better if we had a way to schedule the sync tasks to run on the leader node instead of on the caller node. That would automatically load balance the tasks and would remove the extra rpc call and we would know that the task would always be run on the *same* node which, simplifies things a bit. It would also simplify the code and make things a bit more robust if the cluster topology changes during the queue repair operation...

larshesel:
I see I also need to fix a couple of tests which relied on metadata events to start queues... will be fixed in the next commit

larshesel:
> I see I also need to fix a couple of tests which relied on metadata events to start queues... will be fixed in the next commit

Fixed the `vmq_upgrade_SUITE` test suite.

larshesel:
Fixed review stuff and rebased - can you review and maybe approve?

PR #1049, opened 2019-01-24
Add reauthorize subscriptions feature

dergraf:


rebar.lock
```diff
@@ -102,7 +102,7 @@
  {<<"unicode_util_compat">>,{pkg,<<"unicode_util_compat">>,<<"0.2.0">>},2},
  {<<"vernemq_dev">>,
   {git,"git://github.com/erlio/vernemq_dev.git",
-       {ref,"12e00f7cfc17f067470f3c690e2536b0c68ecf21"}},
```
larshesel: Could you change `rebar.config` to point to `github.com/vernemq/vernemq_dev.git` now you're at it?

Issue #1045, opened 2019-01-22
Client terminated with reason: no case clause matching #vmq_msg{}


boboxiaodd:
### Environment

- VerneMQ Version: 1.6.2
- OS: centos7
- Erlang/OTP version (if building from source):
- VerneMQ configuration (vernemq.conf) or the changes from the default
- Cluster size/standalone:

### Expected behavior
error logger:
```erlang
2019-01-21 20:09:00.013 [error] <0.29450.2>@vmq_mqtt_fsm:connected:334 Ranch listener {{0,0,0,0},9000} terminated with reason: no case clause matching {vmq_msg,<<55,217,248,92,235,34,154,74,102,128,125,245,72,149,69,45>>,[<<"u_70000026">>,<<"message">>],<<...>>,...} in vmq_mqtt_fsm:connected/2 line 334
```

i check source ：
```erlang
connected(#mqtt_pubcomp{message_id=MessageId}, State) ->
    #state{waiting_acks=WAcks} = State,
    %% qos2 flow
    _ = vmq_metrics:incr_mqtt_pubcomp_received(),
    case maps:get(MessageId, WAcks, not_found) of
        #mqtt_pubrel{} ->
            handle_waiting_msgs(State#state{
                                  waiting_acks=maps:remove(MessageId, WAcks)});
        not_found -> % error or wrong waiting_ack, definitely not well behaving client
            lager:debug("stopped connected session, due to qos2 pubrel missing ~p", [MessageId]),
            _ = vmq_metrics:incr_mqtt_error_invalid_pubcomp(),
            terminate(normal, State)
    end;
```

i don't konw why `WAcks` matched `#vmq_msg{}`

dergraf:
Can you reproduce this with this particular client id?

boboxiaodd:
maybe client send pubcomp delay ,and other msg come in ?

boboxiaodd:
```erlang
10:35:22.881 [info] handle_messages:873 deliver_pubrel waiting_acks:#{1 => {mqtt_pubrel,1}}

10:35:22.882 [info] prepare_frame:918 prepare_frame waiting_acks:#{1 => {vmq_msg,<<154,188,21,166,222,10,84,97,119,130,190,84,62,229,53,120>>,[<<"u_16980398">>,<<"message">>],<<"msg payload">>}}

[...]

PR #1053, opened 2019-01-25
Fix bug: waiting_acks add a msg but didn't increment next_msg_id 

boboxiaodd:
https://github.com/vernemq/vernemq/issues/1045


dergraf:
@larshesel I think I understand the bug now, a retried PUBREL after a client reconnect can conflict with a message id of a 'normal' deliver.

@boboxiaodd thanks for your PR, but incrementing the message id isn't enough. We'll look at it soon.

PR #1054, opened 2019-01-25
Fix SWC read-your-own-write semantics

dergraf:


apps/vmq_swc/src/vmq_swc_store.erl
```diff
@@ -476,6 +477,23 @@ get_cached_node_clock(#swc_config{store=StoreName}) ->
...
+            Obj
+    end.
+
+r_o_w_cache_insert_object(#swc_config{store=StoreName}, SKey, Obj) ->
+    ets:insert(StoreName, {SKey, Obj}).
+
+r_o_w_cache_clear(#swc_config{store=StoreName}) ->
+    % the table max size is the size of a write batch, a select_delete shouldn't harm
+    % the overall performance
+    ets:select_delete(StoreName, [{{'$1', '_'}, [{'=/=', '$1', node_clock}], [true]}]),
```
larshesel: This is to me a sign that we should not use the same table. If we later decided to add something else to the table this would delete that something

dergraf: good point...

PR #1065, opened 2019-01-31
Delete cluster state on node which is leaving the cluster

larshesel:
This ensures we don't end up in the node rename case in `vmq_server_sup` and also that the node won't try to reconnect with the remaining nodes. The subscriptions from other nodes are still there, of course, taking space - with swc that information would eventually be removed.


larshesel:
> Oh wow, that was easy. 1st_place_medal

hehe, yeah - do you know/remember why we didn't bring that line over from the original plumtree implementation? I have a feeling there must have been a reason?

PR #1076, opened 2019-02-14
Use colored log output for lager console backend

dergraf:


larshesel:
"I officially renounce the search for the newer and shinier." - https://blog.codinghorror.com/the-magpie-developer/

dergraf:
Talked to one of the K8s experts... 
my questions was:
> Do you know if it is problematic for Kubernetes ( and used logging solutions like stern) if the STDOUT uses color coded loglines? 

Response I got:

> It's apparently a problem in the kubernetes dashboard but I don't know if they fixed that. Apparently stern strips color codes I'm guessing you're pretty safe.

stern is one of the 'cool' K8s log tailing solutions.

Maybe @drf and has some opinions on that?

PR #1078, opened 2019-02-18
Fix outgoing qos2 message id bug (#1045)

### LINKED ISSUE #1045: Client terminated with reason: no case clause matching #vmq_msg{}

larshesel:
Fixes #1045 

@dergraf note my assumption about `pub_rel`s being send to the session before any normal publishes (without message ids) - I hope this is correct.

### DIFF
--- a/apps/vmq_server/src/vmq_mqtt5_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt5_fsm.erl
@@ -901,7 +901,8 @@ register_subscriber(#mqtt5_connect{username=User}=F, OutProps0,
     CoordinateRegs = maps:get(coordinate_registrations, DOpts, ?COORDINATE_REGISTRATIONS),
     case vmq_reg:register_subscriber(CAPSettings#cap_settings.allow_register, CoordinateRegs, SubscriberId, CleanStart, QueueOpts) of
         {ok, #{session_present := SessionPresent,
-               initial_msg_id := MsgId}, QPid} ->
+               initial_msg_id := MsgId,
+               queue_pid := QPid}} ->
             monitor(process, QPid),
             _ = vmq_plugin:all(on_register_m5, [Peer, SubscriberId,
                                                 User, OutProps0]),
--- a/apps/vmq_server/src/vmq_mqtt5_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt5_fsm.erl
@@ -54,7 +54,7 @@
 
 -record(state, {
           %% mqtt layer requirements
-          next_msg_id=1                     :: msg_id(),
+          next_msg_id=undefined             :: undefined | msg_id(),
           subscriber_id                     :: undefined | subscriber_id() | {mountpoint(), undefined},
           will_msg                          :: undefined | msg(),
           waiting_acks=maps:new()           :: map(),
@@ -900,12 +900,14 @@ register_subscriber(#mqtt5_connect{username=User}=F, OutProps0,
                                       def_opts=DOpts} = State) ->
     CoordinateRegs = maps:get(coordinate_registrations, DOpts, ?COORDINATE_REGISTRATIONS),
     case vmq_reg:register_subscriber(CAPSettings#cap_settings.allow_register, CoordinateRegs, SubscriberId, CleanStart, QueueOpts) of
-        {ok, SessionPresent, QPid} ->
+        {ok, #{session_present := SessionPresent,
+               initial_msg_id := MsgId}, QPid} ->
             monitor(process, QPid),
             _ = vmq_plugin:all(on_register_m5, [Peer, SubscriberId,
                                                 User, OutProps0]),
             OutProps1 = maybe_set_receive_maximum(OutProps0, ReceiveMax),
-            check_will(F, SessionPresent, OutProps1, State#state{queue_pid=QPid,username=User});
+            check_will(F, SessionPresent, OutProps1,
+                       State#state{queue_pid=QPid,username=User, next_msg_id=MsgId});
         {error, Reason} ->
             lager:warning("can't register client ~p with username ~p due to ~p",
                           [SubscriberId, User, Reason]),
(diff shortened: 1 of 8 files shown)


apps/vmq_server/src/vmq_mqtt5_fsm.erl
```diff
@@ -1376,11 +1376,19 @@ handle_messages([{deliver, QoS, Msg} = Obj|Rest], Frames, PubCnt, State, Waiting
...
     %% this is called when a pubrel is retried after a client reconnects
     #state{waiting_acks=WAcks} = State0,
     _ = vmq_metrics:incr(?MQTT5_PUBREL_SENT),
-    State1 = State0#state{waiting_acks=maps:put(MsgId, Frame, WAcks)},
+    %% Assumption: all `deliver_pubrel` messages are delivered from
+    %% the queue when a session is added *before* any other messages.
+    %% The pubrel messages has fixed message-ids and we therefore need
+    %% ensure the first message id assigned to a publish message is
+    %% larger than any of the pubrel messages to avoid conflicts.
+    NextMsgId = lists:max([MsgId + 1, CurrNextMsgId]),
```
dergraf: The `MsgId + 1` could theoretically overrun the message id limit `max(ushort)`.

apps/vmq_server/test/vmq_publish_SUITE.erl
```diff
@@ -459,6 +460,87 @@ publish_c2b_retry_qos2_test(_Config) ->
     ok = expect_alive(Socket),
     ok = gen_tcp:close(Socket).
 
+publish_b2c_ensure_valid_msg_ids_test(Config) ->
```
dergraf: it would be interesting to test what happens if multiple of such PUBRELs were persisted.

apps/vmq_server/src/vmq_mqtt5_fsm.erl
larshesel: Good point.. Will fix that.

apps/vmq_server/test/vmq_publish_SUITE.erl
larshesel: Which test scenario are you thinking about here? I agree more cases could/should be covered, ideally we should property test stuff like this, if possible.

larshesel:
@dergraf I reverted the incorrect fix and implemented a new one where we store the `next_msg_id` in the queue. This, however, revealed another problem (the `in_order_delivery_SUITE` started failing): namely that we do not remember the `msg_id` on outgoing qos2 messages (which didn't yet receive a `pubrec`) when we store waiting acks - so when the client comes back the outgoing publish will set the dup flag to ture and assign a new `msg_id` which is against the spec. So next step is to also remember the msgid used when sending the initial qos2 publish message out when storing the unacked stuff in the queue.

apps/vmq_server/src/vmq_mqtt_fsm.erl
```diff
@@ -539,18 +539,20 @@ check_user(#mqtt_connect{username=User, password=Password} = F, State) ->
         false ->
             case auth_on_register(User, Password, State) of
                 {ok, QueueOpts, #state{peer=Peer, subscriber_id=SubscriberId, clean_session=CleanSession,
-                                       cap_settings=CAPSettings, def_opts=DOpts} = NewState} ->
+                                       cap_settings=CAPSettings, def_opts=DOpts} = State1} ->
                     CoordinateRegs = maps:get(coordinate_registrations, DOpts, ?COORDINATE_REGISTRATIONS),
                     case vmq_reg:register_subscriber(CAPSettings#cap_settings.allow_register, CoordinateRegs, SubscriberId, CleanSession, QueueOpts) of
-                        {ok, SessionPresent, QPid} ->
+                        {ok, #{session_present := SessionPresent,
+                               initial_msg_id := MsgId}, QPid} ->
```
dergraf: We could return the `QPid` also as part of the map.

larshesel: yeah, I guess that makes sense. Then spec it as a mandatory field in the map.

larshesel:
Since last I:

 - moved the queue pid into the map returned by `vmq_reg:register_subscriber/4`
 - changed the `{deliver, QoS, Msg}` tuple into a record `#deliver{msg_id, qos, msg}` and use this everywhere *except* when enqueueing as we do this across nodes. For some scenarious this means I have to translate between the two, for example in `vmq_queue:enqueue_many` and when draining.

larshesel:
Had to rebase to get CI to run...

larshesel:
@boboxiaodd, I think this is about done - would you be able to test it out?

boboxiaodd:
@larshesel , Great, i clone the master  and  test it now!

PR #1081, opened 2019-02-19
Fix `vmq_diversity` cache invalidation timing issue (#996).

### LINKED ISSUE #996: Last Will (LWT) authentication error.
### Environment
- VerneMQ Version: 1.6.2
- OS: Official Docker image.
- VerneMQ configuration: Disallow anonymous and activate diversity and postgres plugin.

### Expected behavior
When I connect a new client with clientId and username which previously has connected with last will testament, and not clossed connection, the last will payload is posted in the last will topic without any authentication log warning.

### Actual behaviour
When I do which I explain above, I get the following errors:
[error] can't auth publish [<<"xxx">>,{[],<<"xxx">>},0,[<<"xxx">>,<<"x">>,<<"xxx">>],<<"xxx">>,false] due to not_authorized
[warning] can't authenticate last will for client {[],<<"xxx">>} due to not_allowed.

The postgres ACL and authentication is working as expected, the client has permission to publish in the last will topic, even the last will message is published as expected, but the warning is annoyoing.

Maybe it is a simple log bug or maybe I missed some configuration, but If you could check it, I will be very grateful.

Thank you :+1:

larshesel:
This should fix #996 

The tradeoff is that the cache table is now an `ordered_set` table and lookups are now `ets:select` based on `{SubscriberId, '_'}` which is proportional to the number of entries with that subscriber id modulo the O factor from the ordered set. So basically O(log n) where n is the number of different subscribers.

### DIFF
--- a/apps/vmq_diversity/src/vmq_diversity_cache.erl
+++ b/apps/vmq_diversity/src/vmq_diversity_cache.erl
@@ -18,6 +18,8 @@
 -dialyzer(no_undefined_callbacks).
 -behaviour(gen_server2).
 
+-include_lib("stdlib/include/ms_transform.hrl").
+
 %% API
 -export([start_link/0,
          install/1,
@@ -40,7 +42,11 @@
 
 -import(luerl_lib, [badarg_error/3]).
 
--record(state, {}).
+-record(state,
+        {
+         %% logical timestamp
+         lts = 0 :: non_neg_integer()
+        }).
 
 
 -record(publish_acl, {
@@ -55,6 +61,13 @@
           max_qos = 2,
           modifiers
          }).
+
+-define(ms(Key),
+        %% Key = {<<>>, <<"clientid">>},
+        %% ets:fun2ms(fun({{AKey, '_'}, '_', '_'}=A) when AKey =:= Key -> A end).
+        [{{{'$1','_'},'_','_'},
+          [{'=:=','$1',{const,Key}}],
+          ['$_']}]).
 
 %%%===================================================================
 %%% API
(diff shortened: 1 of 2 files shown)


apps/vmq_diversity/src/vmq_diversity_cache.erl
```diff
@@ -346,12 +364,22 @@ key(MP, ClientId) when is_binary(MP)
...
     Type = type(Input),
-    case ets:lookup(table(cache), Key) of
+    case ets:select(table(cache), ?ms(Key)) of
         [] -> no_cache;
         [{_, PubAclHashes, _}] when Type == publish ->
             match_(table(Type), Input, PubAclHashes);
         [{_, _, SubAclHashes}] when Type == subscribe ->
-            match_(table(Type), Input, SubAclHashes)
+            match_(table(Type), Input, SubAclHashes);
+        Elems ->
```
dergraf: as this is the generalization of the two case clauses above, I wonder if we should replace them by the generalization. I guess lists:reverse on a one-element list is free.

apps/vmq_diversity/src/vmq_diversity_cache.erl
```diff
@@ -18,6 +18,8 @@
 -dialyzer(no_undefined_callbacks).
 -behaviour(gen_server2).
 
+-include_lib("stdlib/include/ms_transform.hrl").
```
dergraf: Not required, correct?

apps/vmq_diversity/src/vmq_diversity_cache.erl
dergraf: Alternatively, one could use `ets:select_reverse`  instead of the extra `lists:reverse`

apps/vmq_diversity/src/vmq_diversity_cache.erl
larshesel: yep, removed

apps/vmq_diversity/src/vmq_diversity_cache.erl
larshesel: Right, I wasn't thinking it through when trying to make this efficient. Since `ets:select_reverse` exists (didn't know that, thanks), I simplified this a bit further, as I could simply ask for a single element.

Issue #1080, opened 2019-02-19
Lua plugin issue: json.decode() throws an error on empty json


ramikushnir:
Hi, I'm experiencing an issue with the lua plugin, tried to look for it in the older issues but didn't find it or issues related to this problem.

### Environment

- VerneMQ Version: 1.6.1
- OS: K8S + Container OS is Debian Jessie
- Cluster size/standalone: 3 nodes

### Expected behavior
In a lua script, when calling json.decode with an empty json, e.g. 
```
json.decode('{}'} 
```
or 
```
json.decode('{"something":{}}')
```
the result should be a table. 

### Actual behaviour
When calling `json.decode()` with an empty json, or an empty json property, the script throws an error and exits.

I've added a piece of code right after my `auth_on_register` implementation, to simulate this.

code: 
```
-- the function that implements the auth_on_register/5 hook
-- the reg object contains everything required to authenticate a client
function my_auth_on_register(reg)

    local my_empty_json="{}";
    log.debug('About to decode json: ' .. my_empty_json);
    local decoded_table = json.decode(my_empty_json);
    log.debug('This line should be printed after the empty json was decoded.');
...
```

This is the log I am seeing after tring to connect to the broker:
```
15:06:23.486 [debug] 0ms mailbox traversal, schedule next lazy broadcast in 10000ms, the min interval is 10000ms
15:06:24.009 [debug] About to decode json: {}
15:06:24.052 [error] can't call function auth_on_register with args [{addr,<<"172.17.0.1">>},{port,46438},{mountpoint,<<>>},{client_id,<<"my-client-1">>},{username,nil},{password,nil},{clean_session,true}] in "/tmp/vmq_scripts/vmq_auth_plugin.lua" due to {error,badarg}
15:06:24.052 [warning] can't authenticate client {[],<<"my-client-1">>} due to lua_script_error
15:06:24.054 [debug] session normally stopped
```

I've tried this also with JSON strigns that contain empty properties, like `{"myProperty":{}}` and I'm getting the same error.

When testing this code with a lua 5.1 engine, not in the vernemq environment it works fine.

Thanks!

larshesel:
Thanks for reporting. We'll look into it.

dergraf:
Fixed in current master

PR #1082, opened 2019-02-20
Fix empty Json Object conversion in Lua scripts

dergraf:


dergraf:
Fixes #1080

apps/vmq_diversity/test/json_test.lua
```diff
@@ -37,3 +37,12 @@ c2 = "{\"a\":{\"b\":{\"c\":\"hello\",\"d\":[1.0,2.0,3.0,4.0,5.0]}}}"
...
 
+d1 = {}
+d2 = "{}"
+assert(equals(d1, json.decode(d2)))
+assert(equals(json.encode(d1), d2))
+
+e1 = {a = {{b = {}}}}
+e2 = "{\"a\":[{\"b\":{}}]}"
+assert(equals(e1, json.decode(e2)))
+assert(equals(json.encode(e1), e2))
```
larshesel: So something like `"{\"a\":[]}"` would decode to `{a = {}}`, and would therefore not be "reversible" (`x =/= decode(encode(x))`. Wouldn't it be good to make this behaviour explicit by adding a test case with it?

larshesel: And maybe even add a comment that it's on purpose?

Issue #1084, opened 2019-02-22
Problems bridging VerneMQ to AWS IoT


Fettuccine-Alfredo:
### Environment

- VerneMQ Version: 1.7.0
- OS: Fedora 29 running docker-vernemq 1.7.0
- Erlang/OTP version (if building from source):
- VerneMQ configuration (vernemq.conf) or the changes from the default
[vernemq.conf.txt](https://github.com/vernemq/vernemq/files/2892818/vernemq.conf.txt)
- Cluster size/standalone: standalone

### Expected behavior

- Expected bridging to AWS IoT to work.

### Actual behaviour

- After each message is bridged/published to AWS IoT, the VerneMQ client disconnects.  After a few cycles of connect/disconnect, AWS IoT refused to accept new connections.

Here's the log output: [vernemq.log.txt](https://github.com/vernemq/vernemq/files/2892828/vernemq.log.txt)

dergraf:
The no case clause match error is here: https://github.com/vernemq/vernemq/blob/master/apps/vmq_commons/src/gen_mqtt_client.erl#L455. But instead of the `#mqtt_publish` it tries to match a tuple likely coming from here https://github.com/vernemq/vernemq/blob/master/apps/vmq_commons/src/gen_mqtt_client.erl#L605

dergraf:
@larshesel bug identified, working on a fix

Fettuccine-Alfredo:
The crash has gone away, but the vmq_bridge no longer attempts to send the message to AWS IoT:

01:26:48.679 [warning] dropped message {deliver,0,undefined,{vmq_msg,<<70,185,60,216,186,180,209,154,199,73,165,184,90,181,145,230>>,[<<"test">>],<<"test2">>,false,false,0,[],false,prefer_local,#{},undefined}} for plugin vmq_bridge

larshesel:
Hi @Fettuccine-Alfredo I think this is due to an otherwise unrelated refactoring I was doing - I'll investigate and get back later today. Sorry for the inconvenience.

larshesel:
Hi @Fettuccine-Alfredo I created a PR (#1087) to address this problem - it's in review now.

larshesel:
Hi @Fettuccine-Alfredo the PR #1087 was merged, it would be great if you could retry the bridge and report if it works! Thanks!

Fettuccine-Alfredo:
Seems to be working now.

larshesel:
[...]

PR #1085, opened 2019-02-22
Fix retry of MQTT publish frame in vmq-bridge

dergraf:
Fixes https://github.com/vernemq/vernemq/issues/1084


Fettuccine-Alfredo:
The crash has gone away, but the vmq_bridge no longer attempts to send the message to AWS IoT:

01:26:48.679 [warning] dropped message {deliver,0,undefined,{vmq_msg,<<70,185,60,216,186,180,209,154,199,73,165,184,90,181,145,230>>,[<<"test">>],<<"test2">>,false,false,0,[],false,prefer_local,#{},undefined}} for plugin vmq_bridge

PR #1095, opened 2019-02-28
Use atomics instead of mzmetrics

larshesel:
This replaces the use of the `mzmetrics` counter library with the built-in `atomics` library (as of Erlang/OTP 21.2).

Further, when updating a counter the `persistent_term` library is used to look up the reference to the `atomics` array and each metric is now assigned an specific index in this array. Caller processes cache the `atomics` array reference, so it doesn't have to lookup the value in the `persistent_term` store. Before using `mzmetrics` each metric used would cause the `mzmetrics` reference to be stored.

This implementation seems to be about %27 faster on average than the `mzmetrics` one.


larshesel:
Added the extra allocations as discussed and only allocate a new array if one didn't already exist

PR #1100, opened 2019-03-04
`vmq_reg:publish/4` returns the number of matching local and remote subscriptions

dergraf:


changelog.md
```diff
@@ -19,6 +19,9 @@
 - Fix retry of MQTT publish frame in `vmq_bridge` (#1084).
 - Fix outgoing qos2 message id bug (#1045).
 - VerneMQ now requires Erlang/OTP 21.2 or newer to build.
+- Change `vmq_reg:publish/4` to return the number of matched subscriptions. This
```
codeadict: Could you explain how it breaks that publish API and how it could be used instead?

larshesel: as far as I understood it it is only the return value which is different. Before `ok` was returned, now `{ok, {NumLocalMatches,NumRemoteMatches}}` is returned. Maybe it would be good to mention this in the changelog... /cc @dergraf

dergraf: Done

codeadict: Looks good to me, any specific reason why this Api is not made public as part of vernemq_dev? That would be  a nice way to track changes, even if a major change is done people will easily know about it because is a public/documented Api IMHO. Pretty sure lot of people use it in their plugins as i do myself, i find this Api to be probably one of the most powerful functionalities of VerneMQ against other broker we tried.

changelog.md
```diff
@@ -21,6 +21,11 @@
 - VerneMQ now requires Erlang/OTP 21.2 or newer to build.
 - Use `atomics` and `persistent_term` to handle metrics instead of using the
   `mzmetrics` library which has been removed.
+- Change `vmq_reg:publish/4` to return `{ok, {NumLocalMatches, NumRemoteMatchs}}`
+  instead of `ok`. This introduces a breaking change to the *inofficial* publish
```
codeadict: Although the Merriam-Webster’s dictionary mentions `inofficial` as a synonym, `unofficial` is the most common method that English speaking lads use.

PR #1105, opened 2019-03-09
Fix remove unnecessary app startup

dergraf:


rebar.config.script
```diff
@@ -0,0 +1,28 @@
+{relx, RelxConfig0} = lists:keyfind(relx, 1, CONFIG),
+{release, Version, Apps0} = lists:keyfind(release, 1, RelxConfig0),
+{ok, [{application, vmq_server, VmqServerTerms}]} 
+= file:consult(filename:join(["apps", "vmq_server", "src", "vmq_server.app.src"])),
+DefaultApps = [A ||A <- Apps0, is_atom(A)] ++ proplists:get_value(applications, VmqServerTerms, []),
+
+LoadedAppsFun = fun(F, Type, [App|Rest], Acc, Excl) ->
+                        case lists:keymember(App, 1, Excl) of
+                            false ->
+                                AppSrcFile = filename:join(["_build", "default", "lib", App, src, atom_to_list(App) ++ ".app.src"]),
```
larshesel: is the current profile available from the config? If so I think that should be used instead of `default` here to make sure the code works with any profile.

PR #1120, opened 2019-03-21
Fix new dialyzer errors uncovered by OTP 21.3

larshesel:
Erlang 21.3 has some dialyzer improvements which uncovered a few new issues.


larshesel:
This fixes #1156

PR #1141, opened 2019-04-11
Make MySQL hash_method use salt

Androbin:
The PostgreSQL backend uses salt by default.
This PR makes the MySQL backend use salt as well.


Androbin:
If the lack of customization bothers you just as much as me, maybe take a look at #1142 instead.

Androbin:
@larshesel
* The enum could have separate `_salt` versions instead.
* The string could match the existing values separately.

ioolkos:
This PR got obviously forgotten somehow. I'll take care & decide what to do about it.

PR #1142, opened 2019-04-11
Make MySQL hash_method use string instead of enum

Androbin:
This PR arose as a replacement for #1141.
The enum is unnecessary and uneasy to customize.
All it does is map to string anyway.

The PostgreSQL backend uses salt by default.
Using a string right away makes it trivial to do the same with MySQL.

### DIFF
--- a/apps/vmq_diversity/src/vmq_diversity_mysql.erl
+++ b/apps/vmq_diversity/src/vmq_diversity_mysql.erl
@@ -120,4 +120,11 @@ hash_method(_, St) ->
     {ok, DBConfigs} = application:get_env(vmq_diversity, db_config),
     DefaultConf = proplists:get_value(mysql, DBConfigs),
     HashMethod = proplists:get_value(password_hash_method, DefaultConf),
-    {[HashMethod], St}.
+    MysqlFunc = case HashMethod of
+        password -> <<"PASSWORD(?)">>;
+        md5 -> <<"MD5(?)">>;
+        sha1 -> <<"SHA1(?)">>;
+        sha256 -> <<"SHA2(?, 256)">>;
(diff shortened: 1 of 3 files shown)


apps/vmq_diversity/src/vmq_diversity_mysql.erl
```diff
@@ -124,6 +124,7 @@ hash_method(_, St) ->
         password -> <<"PASSWORD(?)">>;
         md5 -> <<"MD5(?)">>;
         sha1 -> <<"SHA1(?)">>;
-        sha256 -> <<"SHA2(?, 256)">>
-    end, 
+        sha256 -> <<"SHA2(?, 256)">>;
+        _ -> HashMethod
```
Androbin: @larshesel Can you tell if it's `HashMethod` or `<<HashMethod>>` ? I ain't exactly fluid in Erlang.

dergraf: `HashMethod` would be correct. But I don't think you'd need to change the `password`, `md5`, `sha1` and `sha256` to actual strings instead of atoms.

Androbin: @dergraf Excuse me. Did you mean the keys **do** or **do not** need to be quoted?

codeadict: I do agree, atoms should be kept as an enum field, having the user enter free text with no validation for this setting could result in errors and lots of support requests,

larshesel:
Yes, we shouldn't use strings instead of enums. I do think we should investigate using salts, so let's keep the other PR open for now.

PR #1155, opened 2019-04-24
Add TLS support to the vmq_diversity database drivers

### LINKED ISSUE #811: PostgreSQL auth doesn't seem to support SSL connections
### Environment

- VerneMQ Version: 1.5.0
- OS: Docker/Kubernetes

### Expected behavior

VerneMQ should be able to connect to a PostgreSQL server that requires SSL.

### Actual behaviour

There's no way to specify the SSL mode for the connection (_e.g._ "prefer" or "require"), and connecting to a PostgreSQL server that requires SSL fails with:
```
[warning] Could not connect to postgresql due to invalid_authorization_specification
```
(repeated at a high rate in the logs)

The error message is a bit vague and I'm not 100% sure it fails because of SSL, but it seems likely since my Postgres instance requires SSL and I checked multiple times that I'm using the correct connection info.

Also I quickly checked the code and it's not setting the ssl option when connecting with epgsql (at https://github.com/erlio/vernemq/blob/1.5.0/apps/vmq_diversity/src/vmq_diversity_sup.erl#L87), so it seems to confirm my suspicions :)

larshesel:
This adds basic TLS support for the postgresql driver. This is also
needed in order to use the posgresql driver to connect to a secure
CockroachDB installation.

- [x] postgresql
- [x] mongdb
- [ ] mysql (afaict emysql doesn't support TLS)
- [ ] redis (doesn't support TLS)

Fixes #811

### DIFF
--- a/apps/vmq_diversity/priv/vmq_diversity.schema
+++ b/apps/vmq_diversity/priv/vmq_diversity.schema
@@ -208,6 +208,40 @@
   {default, safe},
   hidden]}.
 
+%% @doc Specify if the mongodb driver should use TLS or not.
+{mapping, "vmq_diversity.mongodb.ssl", "vmq_diversity.db_config.mongodb.ssl",
+ [{datatype, flag},
+  {default, off}
+ ]}.
+
+%% @doc The cafile is used to define the path to a file containing
+%% the PEM encoded CA certificates that are trusted.
+{mapping, "vmq_diversity.mongodb.cafile", "vmq_diversity.db_config.mongodb.cacertfile",
+ [
+  {default, ""},
+  {datatype, file},
+  {commented, "{{platform_etc_dir}}/cafile.pem"},
+  {validators, ["file-exists", "file-is-readable"]}
+ ]}.
+
+%% @doc Set the path to the PEM encoded server certificate.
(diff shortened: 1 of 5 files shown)


apps/vmq_diversity/priv/vmq_diversity.schema
```diff
@@ -70,6 +70,40 @@
   {default, 5},
   hidden]}.
 
+%% @doc Specify if the postgresql driver should use TLS or not.
+{mapping, "vmq_diversity.postgres.ssl", "vmq_diversity.db_config.postgres.ssl",
```
dergraf: should we call it tls?

codeadict: Calling it TLS is probably better, ssl is a replaced protocol. People still keep calling ssl to any encrypted TCP connection because they are used to the term

larshesel: I kept the SSL name as that's what we use everywhere else at the moment. So what do you think? Should we change this to TLS now and then the rest in VMQ 2.0 or shall we do all of them for VMQ 2.0 and keep SSL for now?

dergraf: Let's keep SSL

PR #1160, opened 2019-05-02
Use bcrypt on multiple cores

### LINKED ISSUE #579: Run bcrypt on multiple cores [was: Authentication with Redis fails "due to error"]
### Environment

- VerneMQ Version: 1.2.3
- OS: Docker

### Configuration

```
  mqtt:
    environment:
      - DOCKER_VERNEMQ_LISTENER__MAX_CONNECTIONS=96000
      - DOCKER_VERNEMQ_LISTENER__NR_OF_ACCEPTORS=100
      - DOCKER_VERNEMQ_MAX_CLIENT_ID_SIZE=32
      - DOCKER_VERNEMQ_MAX_MESSAGE_SIZE=524288
      - DOCKER_VERNEMQ_PERSISTENT_CLIENT_EXPIRATION=1h
      - DOCKER_VERNEMQ_PLUGINS__VMQ_ACL=off
      - DOCKER_VERNEMQ_PLUGINS__VMQ_DIVERSITY=on
      - DOCKER_VERNEMQ_PLUGINS__VMQ_PASSWD=off
      - DOCKER_VERNEMQ_VMQ_DIVERSITY__AUTH_REDIS__ENABLED=on
      - DOCKER_VERNEMQ_VMQ_DIVERSITY__REDIS__HOST=redis
    image: erlio/docker-vernemq:1.2.3
    ports:
      - "8081:8080"
      - "8888:8888"
    restart: always
    ulimits:
      nofile:
        soft: 100000
        hard: 100000
```

### Problem

I have a single client that – after working fine for weeks – can no longer connect. Every attempt results in a CONNACK 5 (unauthorized). I triple-checked the password and manually validated the bcrypt hash in redis.

The only log message I get is:

```
[warning] <0.3218.0>@vmq_mqtt_fsm:check_user:544 can't authenticate client {[],<<"XXX">>} due to error
```

Peeking at the code, the word "error" is formatted into that log line, in place of a more informative description of the problem. It originates somewhere deeper down the stack.

Of course, this only occurs in production. All other clients work just fine and I can't reproduce the problem elsewhere. The only thing that is special about this client is that it may have had a higher volumes of messages in general and misbehaved in the past (excessive reconnects, simultaneous/duplicate connections).

With my very limited understanding of Erlang and VerneMQ, I was unable to figure out where that "error" originated. Without going deeper into why this single client acts up, maybe we can first figure out what the actual error is and where it originates? Better error messages are always a good thing :)

P.S.: This may or may not be related, but I have seen this error sporadically for other clients too:

```
[error] <0.725.0>@vmq_diversity_script:call_function:64 can't call into Lua sandbox for function auth_on_register due to {timeout,{gen_server,call,[<0.601.0>,{call_function,auth_on_register,[{addr,<<"172.18.0.1">>},{port,58256},{mountpoint,<<>>},{client_id,<<"XXX">>},{username,<<"XXX">>},{password,<<"XXX">>},{clean_session,true}]}]}}
```

That doesn't appear to be a persistent problem however.

larshesel:
- [x] add a vmq_bcrypt.pool_size schema setting
- [x] allow automatic detection of all cores as the pool size.

Note, if this gets merged we should add a note about the setting in the `vmq_diversity` docs vernemq/vmq-docs#18

Fixes #579

### DIFF
--- a/apps/vmq_diversity/priv/vmq_bcrypt.schema
+++ b/apps/vmq_diversity/priv/vmq_bcrypt.schema
@@ -1,0 +1,14 @@
+%% @doc The pool_size specifies how many bcrypt operations that are
+%% allowed concurrently.
+{mapping, "vmq_bcrypt.pool_size", "bcrypt.pool_size", [
+                                                       {datatype, integer},
+                                                       {default, 1},
+                                                       {include_default, 1}
+                                                      ]}.
+
+{mapping, "vmq_bcrypt.mechanism", "bcrypt.mechanism", [
+                                                       {datatype, {enum, [nif, port]}},
+                                                       {default, port},
+                                                       hidden,
+                                                       {include_default, port}
+                                                      ]}.
--- a/apps/vmq_diversity/priv/vmq_bcrypt.schema
+++ b/apps/vmq_diversity/priv/vmq_bcrypt.schema
@@ -1,16 +1,19 @@
 %% @doc The pool_size specifies how many bcrypt operations that are
-%% allowed concurrently.
+%% allowed concurrently. The value `auto` will try to detect all
+%% logical cpus and set the pool size to that number. If the number of
+%% logical cpus cannot be detected, a value of 1 is used.
 {mapping, "vmq_bcrypt.pool_size", "bcrypt.pool_size", [
-                                                       {datatype, [integer, {enum, [logical_processors_online]}]},
+                                                       {datatype, [integer, {enum, [auto]}]},
                                                        {default, 1},
-                                                       {include_default, 1}
+                                                       {include_default, 1},
+                                                       {validators, ["non-zero-positive-int"]}
                                                       ]}.
 
 {translation,
  "bcrypt.pool_size",
  fun(Conf) ->
          case cuttlefish:conf_get("vmq_bcrypt.pool_size", Conf) of
-             logical_processors_online ->
+             auto ->
                  case erlang:system_info(logical_processors_online) of
                      unknown ->
                          1;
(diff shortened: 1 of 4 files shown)


apps/vmq_diversity/priv/vmq_bcrypt.schema
```diff
@@ -0,0 +1,30 @@
...
+                                                       {datatype, [integer, {enum, [logical_processors_online]}]},
+                                                       {default, 1},
+                                                       {include_default, 1}
+                                                      ]}.
+
+{translation,
+ "bcrypt.pool_size",
+ fun(Conf) ->
+         case cuttlefish:conf_get("vmq_bcrypt.pool_size", Conf) of
+             logical_processors_online ->
```
dergraf: should we call this 'auto'?

larshesel: yeah, let's pick sth which isn't mapped directly to the erlang term. I think `max` would be more descriptive, though.

Btw. I named the setting `vmq_bcrypt.pool_size`, it could also live in the global scope and be just something like `bcrypt_pool_size`. I'm not sure if we have some implicit rules about that?

larshesel: thinking a bit more, `max` isn't ideal either as you can of course have more ports ready than the number of logical cpus - perhaps `auto` isn't so bad after all. I'll change it to `auto`.

dergraf:
i saw you added the new schema file and renamed the exidting ones... Do you know how this will be treated when upgrading e.h. a RPM?

larshesel:
Good question. I would expect the rpm to delete the old files and replace them with the new ones. If the RPM doesn't clean things up then it is IMO broken.

larshesel:
I tested with a debian package and the old `18-vmq_swc.schema` files was removed and in its place `18-vmq_bridge.schema` was copied.

PR #1162, opened 2019-05-02
Debug readable ct output crash

larshesel:
since a while using `rebar3 ct` is broken and we have to pass `--readable=false` to run the tests. This is super annoying as the output is ... unreadable.

Was finally able to reproduce locally and opened an issue on the rebar3 repo: https://github.com/erlang/rebar3/issues/2063

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ jobs:
   include:
     - stage: Test
       before_script: "epmd -daemon"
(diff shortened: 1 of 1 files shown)


larshesel:
There's a PR waiting to be merged for this issue (https://github.com/erlang/rebar3/pull/2065), but this PR works around the issue until the fix is in a new `rebar3` release.

PR #1175, opened 2019-05-11
vmq_pulse

dergraf:
Add the `vmq_pulse` plugin. This plugin periodically sends cluster information to
  a pulse backend via HTTP(S). The plugin is NOT enabled by default, and once
  enabled it has to be setup with `vmq-admin pulse setup`. This enables a support
  crew to analyze the cluster performance without having direct access to the VerneMQ
  nodes. Such a 'Pulse' contains the following information:
  - Cluster status, similar to `vmq-admin cluster show`
  - Plugins, similar to `vmq-admin plugin show --internal`
  - Metrics, similar to `vmq-admin metrics show`
  - Names and versions of loaded OTP applications
  - Erlang System Version
  - OS Kernel information, output of `uname -a`
  - Pulse Version


apps/vmq_pulse/src/vmq_pulse_exporter.erl
```diff
@@ -0,0 +1,206 @@
...
+                                     name => Name,
+                                     desc => Desc,
+                                     value => Val}};
+                       (_UnsupportedMetricDef) ->
+                            false
+                    end, vmq_metrics:metrics(#{aggregate => false}))},
+        {applications, [{App, Vsn} || {App, _, Vsn} <- application:which_applications()]},
+        {system_version, erlang:system_info(system_version)},
+        {cluster_status, vmq_cluster:status()},
+        {uname, os:cmd("uname -a")}
```
larshesel: Can we add system memory and usage as well as disk usage information?

apps/vmq_pulse/src/vmq_pulse_exporter.erl
```diff
@@ -0,0 +1,206 @@
...
+    ],
+    zlib:compress(term_to_binary(Body)).
+
+push_body(PulseUrl, ApiKey, Body) ->
+    ContentType = "application/x-vmq-pulse-1",
```
larshesel: Does the `ContentType` here express the version? Otherwise I think we should add a version pre-fix to the url-path.

apps/vmq_pulse/src/vmq_pulse_exporter.erl
```diff
@@ -0,0 +1,206 @@
...
+                  _ ->
+                      AuthStr = base64:encode_to_string(?PULSE_USER ++ ":" ++ ApiKey),
+                      [{"Authorization", "Basic " ++ AuthStr}]
+              end,
+    {ok, ConnectTimeout} = application:get_env(vmq_pulse, connect_timeout),
+    HTTPOpts =  [{timeout, 60000},
+                 {connect_timeout, ConnectTimeout * 1000},
+                 {autoredirect, true}],
+
+    case httpc:request(post, {PulseUrl, Headers, ContentType, Body}, HTTPOpts, []) of
```
larshesel: Can a `httpc` connection be kept open so we don't have to create it every time we post? Reading the `httpc` man page I think this is the default...

apps/vmq_pulse/src/vmq_pulse_exporter.erl
dergraf: Yep it does.

apps/vmq_pulse/src/vmq_pulse_exporter.erl
```diff
@@ -0,0 +1,209 @@
...
+%%--------------------------------------------------------------------
+handle_info(push, State) ->
+    ClusterId = vmq_pulse:get_cluster_id(),
+    {ok, PulseUrl} = application:get_env(vmq_pulse, url),
+    {ok, Interval} = application:get_env(vmq_pulse, push_interval),
+    ApiKey = application:get_env(vmq_pulse, api_key, undefined),
+    NodeId = atom_to_binary(node(), utf8),
+    case (ClusterId =/= undefined) of
+        true ->
+            Body = create_body(ClusterId, NodeId),
```
larshesel: If create_body fails for some reason (exception), perhaps we should post that to the endpoint so we have the stack-trace right away and don't have to ask the users for that?

PR #1197, opened 2019-05-21
Make compatible with erlang 22.0-1

mhdhdri:


mhdhdri:
config is a bit outdated.
your erlang version is 21.2.

larshesel:
Hi @mhdhdri. Thanks for the pull request. Migrating from `gen_fsm` to `gen_statem` is on our backlog, but isn't strictly necessary for OTP 22 compatibility as the `gen_fsm` is still part of OTP 22. So, we definitely want to migrate to `gen_statem` at some point, but there are a few issues with your pull request, in that it looks like you've simlpy replaced `gen_fsm` with `gen_statem` in the `vmq_queue` module. Unfortunately it's not as simple as that, and it requires a bit more work to rewrite it to use `gen_statem`, for example `gen_statem` doesn't have the functions `send_event/2`, `sync_send_event/3` and others. See the errors in  the build output: https://travis-ci.org/vernemq/vernemq/jobs/535411968

mhdhdri:
@larshesel
Hi again.
I tried installing verneMQ on a machine running OTP 22 and I got errors saying `gen_fsm` doesn't work anymore. So in addition to replacing some `gen_statem`s I replaced some of `gen_fsm`s with `erlang`( including things like changing `gen_fsm:send_event_after` to `erlang:send_after`). The result was installed seccesfully.

larshesel:
Ok, I'm glad you got it to compile, but I'm afraid we can't accept this PR as is as the test-suite isn't able to run due to the issues I pointed out above which will cause VerneMQ to not work correctly. Closing for now.

PR #1214, opened 2019-06-10
Fix bridge to remove the local/remote prefix from the local/remote subscription and add the remote/local prefix for republishing

alisoam:
Local prefix and remote prefix don't have the expected behavior in the current implementation of the bridge.

### DIFF
--- a/apps/vmq_bridge/src/vmq_bridge.erl
+++ b/apps/vmq_bridge/src/vmq_bridge.erl
@@ -140,11 +140,11 @@ handle_info({deliver_remote, Topic, Payload},
     %% publish an incoming message from the remote broker locally if
     %% we have a matching subscription
     lists:foreach(
-      fun({{in, T}, LocalPrefix}) ->
+      fun({{in, T}, {LocalPrefix, RemotePrefix}}) ->
               case match(Topic, T) of
                   true ->
                       % ignore if we're ready or not.
-                      PublishFun(routing_key(LocalPrefix, Topic), Payload, #{});
+                      PublishFun(swap_prefix(RemotePrefix, LocalPrefix, Topic), Payload, #{});
                   false ->
                       ok
               end;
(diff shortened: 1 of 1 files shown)


larshesel:
Hi, thanks for the PR. As @ioolkos says it would be good to understand the problem/issue this fixes, and to that end a description and concrete examples would be much appreciated. Thanks!

alisoam:
Hi,
as described in the Mosquitto broker configuration file, the local prefix and the remote prefix in the bridge is used as follow:
- For incoming topics, the bridge will prepend the pattern with the remote prefix and subscribe to the resulting topic on the remote broker. When a matching incoming message is received, the remote prefix will be removed from the topic and then the local prefix added.
- For outgoing topics, the bridge will prepend the pattern with the local prefix and subscribe to the resulting topic on the local broker. When an outgoing message is processed, the local prefix will be removed from the topic then the remote prefix added.
This behaviour is crucial in order to prevent loops when you want to share all of the topic tree between two brokers.
For example I wanted to use the bridge in order to connect a Rabbitmq cluster to a Vernemq cluster. I added the following lines in the configuration file:
```
#Other bridge configurations
vmq_bridge.tcp.br0.topic.1 in 0 $empty rabbitmq_out
vmq_bridge.tcp.br0.topic.2 out 0 $empty rabbitmq_in
```
and I expected to receive messages in subscribers to the topic "Test/Test1" in the Vernemq cluster when publishing to the topic "rabbitmq_out.Test.Test1" in the Rabbitmq cluster and receive form the topic "rabbimq_in.Test.Test2" in the Rabbitmq cluster when I publish to the topic "Test/Test2" in the Vernemq cluster.
But in the current implementation of the Vernemeq, the bridge only add the local prefix and the remote prefix to the incoming and outgoing topics. This PR tries to make the Vernemq bridge behave like the Mosuitto broker.

larshesel:
As far as I can tell your analysis is correct and we're not handling the prefixes correctly. We'll need a bit more time to review this and also add some test-cases so we can ensure the behaviour is fixed once and for all. Thanks for your patience and thanks again for the PR!

larshesel:
Functionally this looks good to me - I'll see if I can add a few testscases for this on monday. /cc @dergraf

larshesel:
Hi @alisoam I just create a small test for the prefixes here #1225 (built on top of yours) - does that look correct to you? In any case with your code that test passes, without it it doesn't - so I'd say we should merge this (and the testcase). /cc @dergraf

alisoam:
Hi @larshesel,
I think it's correct. Thanks

PR #1218, opened 2019-06-14
Add client keepalive expiration metric

larshesel:
This adds a new metric `client_keepalive_expired` tracking MQTT keepalive violations.

The initial idea was to use the `client_expiry` metrics was is marked as unused and deprecated. It's not clear to me what the original idea behind `client_expiry` was but perhaps it was thought as to track `persistent_client_expiration`? In any case I thought the name  `client_expiry` was too vague so I didn't re-enable this one, but rather added a new one. If we want to add a metric tracking `persistent_client_expiration` events we could call it `persistent_client_expired` or `offline_client_expired`.

### DIFF
--- a/apps/vmq_server/src/vmq_metrics.erl
+++ b/apps/vmq_server/src/vmq_metrics.erl
@@ -71,8 +71,6 @@
          incr_queue_in/1,
          incr_queue_out/1,
 
-         incr_client_expired/0,
-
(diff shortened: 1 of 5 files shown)


dergraf:
Absolutely agree with you. So `incr_client_expired` wasn't even used, I am shocked! ;)

PR #1220, opened 2019-06-18
Make it possible to reload `vmq_webhooks` config from schema format

larshesel:
This PR makes it possible to reload the webhooks config using the config format defined by the schema file. This is required for the current approach taken in the vernemq operator for dynamically reloading plugin configurations.


larshesel:
I'll close this for now as we decided to go another route in the vernemq operator.

PR #1240, opened 2019-07-03
Separate command-line args from `su` options in `check_user`.

### LINKED ISSUE #1239: vmq-admin args passed to su
### Environment
- VerneMQ Version: 1.8.0
- OS: GNU/Linux (Ubuntu)

### Expected behavior
Running `sudo vmq-admin --help` should print usage.

### Actual behaviour
```console
ubuntu@ip-10-1-2-3:~$ sudo vmq-admin --help
Usage: su [options] [LOGIN]

Options:
  -c, --command COMMAND         pass COMMAND to the invoked shell
  -h, --help                    display this help message and exit
  -, -l, --login                make the shell a login shell
  -m, -p,
  --preserve-environment        do not reset environment variables, and
                                keep the same shell
  -s, --shell SHELL             use SHELL instead of the default in passwd
```
This is just the output of `su --help`.

jwgcarlson:
This makes `sudo vmq-admin --help` do something sensible, rather than just exec'ing `su --help`.

Fixes #1239. (Only tested on Linux though.)

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -8,6 +8,7 @@ Johanna Appel
 Dairon Medina Caro (@codeadict)
 craftup (baranga)
 Riccardo Binetti (@rbino)
+Jordan Carlson (@jwgcarlson)
 Justin Crown
 Lars Hesel Christensen
 Mihail Davydenkov
--- a/changelog.md
+++ b/changelog.md
@@ -28,6 +28,7 @@
 - Fix prefix handling in the bridge plugin (vmq_bridge).
 - Strengthen parameter validation in the `bcrypt.hashpw/2` LUA function in
   `vmq_diversity`.
+- Pass arguments correctly to `vmq-admin` when called via `sudo`.
 
 ## VerneMQ 1.8.0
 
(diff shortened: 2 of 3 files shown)


ioolkos:
@jwgcarlson thanks!
@ engineering department: please review :)

dergraf:
Looks good to me... the Bash on Alpine also support the `--`

Some information from https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean-also-known-as-bare-double-dash:
> More precisely, a double dash (--) is used in bash built-in commands and many other commands to signify the end of command options, after which only positional parameters are accepted.

larshesel:
Hi @jwgcarlson I just rebased this and added a commit with an update to the changelog and added you to the THANKS file. Is this fine with you then I'd squash and merge.

jwgcarlson:
Thanks @larshesel, I just got around to looking at this again tonight. I think you may have misplaced the `(@jwgcarlson)` bit in the THANKS file, otherwise looks great.

larshesel:
Hi @jwgcarlson you're right! whoops - I'll fix that and merge!

larshesel:
Merged! Thanks a lot for the contribution! :)

PR #1248, opened 2019-07-09
Fix Grafana metrics

troyanov:
Metrics 'Current MQTT send rate' and 'Current MQTT Receive rate' might not work because of 'Multiple Series Error'.
This happens when you have both mqtt_version="4" and mqtt_version="5".

To fix this error we can simply apply sum() to get a summary rate on all versions.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -33,6 +33,7 @@ Torsten Rehn
 Tristan Sloughter
 Artur Sobierak
 Ivan Subotic
+Anton Troyanov (@troyanov)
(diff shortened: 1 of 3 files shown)


larshesel:
Hi, thank you for the PR - it looks good. I've added a commit with a changelog entry and added you to the THANKS file. Is this OK with you? Then I'll go ahead and squash and merge.

troyanov:
Hi Lars! Yes, thats totally fine.

PR #1252, opened 2019-07-11
Ensure systemd doesn't terminate VerneMQ if it is slow to start

### LINKED ISSUE #1251: Ensure systemd doesn't terminate VerneMQ if it takes a long time to start
### Environment

- VerneMQ Version: 1.8.0
- OS: Centos
- Cluster size/standalone: any

### Expected behavior

Even if VerneMQ takes a long time to start it should be able to start. Currently if there are many messages in the msgstore it can take a significant amount of time for VerneMQ to start.

### Actual behaviour

When starting VerneMQ with `vernemq start` and VerneMQ takes more than 15 seconds to start an error like this is reported:

```
vernemq failed to start within 15 seconds,
see the output of 'vernemq console' for more information.
If you want to wait longer, set the environment variable
WAIT_FOR_ERLANG to the number of seconds to wait.
```

and an return code of 1 is returned which causes systemd to terminate VerneMQ even though it was still starting.

The solution to this would be to add 

```
Environment="WAIT_FOR_ERLANG=3600"
```

to the VerneMQ service file which would match the `TimeoutStartSec=3600`.

Thanks to @francois-travais for discovering and debugging this issue.

larshesel:
fixes #1251

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -34,6 +34,7 @@
   and 1.7.1 respectively. This update also means the PROXY protocol code was
   removed and the PROXY support from cowboy is used instead.
 - Handle retained flag in the bridge plugin (vmq_bridge).
+- Ensure systemd doesn't terminate VerneMQ if it is slow to start.
 
 ## VerneMQ 1.8.0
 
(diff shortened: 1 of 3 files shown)


francois-travais:
Great, looks good to me! :rocket:

PR #1254, opened 2019-07-12
Fix message store startup consistency checking routine

dergraf:
Fix message store startup consistency checking routine which resulted in
  potentially deleting wrong messages in cases where an inconsistency was detected.
  This fix also increases message storage performance during startup as well as
  during normal operation. However, the nature of the bug and the provided solution
  don't enable a simple downgrade path and introduces a backward incompatibility if
  a VerneMQ message store containing offline messages has to be downgraded to 1.8.0.

### DIFF
--- a/apps/vmq_server/src/vmq_lvldb_store.erl
+++ b/apps/vmq_server/src/vmq_lvldb_store.erl
@@ -23,7 +23,8 @@
          msg_store_delete/2,
          msg_store_find/1,
          get_ref/1,
-         refcount/1]).
+         refcount/1,
+         get_state/1]).
 
 -export([msg_store_init_queue_collector/3]).
 
@@ -111,6 +112,9 @@ get_ref(BucketPid) ->
 refcount(MsgRef) ->
     call(MsgRef, {refcount, MsgRef}).
 
+get_state(Pid) ->
+    gen_server:call(Pid, get_state, infinity).
+
 call(Key, Req) ->
     case vmq_lvldb_store_sup:get_bucket_pid(Key) of
         {ok, BucketPid} ->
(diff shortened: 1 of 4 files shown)


changelog.md
```diff
@@ -39,6 +39,12 @@
   connecting to VerneMQ such that the retained bit is kept on messages routed to
   the bridge (Retained as Publish) and messages coming from the bridge are not
   forwarded to the bridge if it has a matching description (No Local).
+- Fix message store startup consistency checking routine which resulted in
+  potentially deleting wrong messages in cases where an inconsistency was detected.
+  This fix also increases message storage performance during startup as well as
+  during normal operation. However, the nature of the bug and the provided solution
+  don't enable a simple downgrade path and introduces a backward incompatibility if
+  a VerneMQ message store containing offline messages has to be downgraded to 1.8.0.
```
larshesel: to 1.8.0 or earlier versions.

apps/vmq_server/src/vmq_lvldb_store.erl
```diff
@@ -378,20 +375,17 @@ handle_req({read, {MP, _} = SubscriberId, MsgRef},
...
             %% last one to be deleted
-            eleveldb:write(Bucket, [{delete, RefKey},
-                                    {delete, IdxKey},
+            eleveldb:write(Bucket, [{delete, IdxKey},
                                     {delete, MsgKey}], WriteOpts);
         _ ->
             %% we have to keep the message, but can delete the ref and idx
-            eleveldb:write(Bucket, [{delete, RefKey},
-                                    {delete, IdxKey}], WriteOpts)
+            eleveldb:write(Bucket, [{delete, IdxKey}], WriteOpts)
```
larshesel: The comment above should no longer mention `ref`

apps/vmq_server/src/vmq_lvldb_store.erl
```diff
@@ -341,20 +341,17 @@ handle_req({write, {MP, _} = SubscriberId,
...
             %% new message
             Val = serialize_p_msg_val_pre({RoutingKey, Payload}),
             eleveldb:write(Bucket, [{put, MsgKey, Val},
-                                    {put, RefKey, <<>>},
                                     {put, IdxKey, IdxVal}], WriteOpts);
         _ ->
             %% only write ref
-            eleveldb:write(Bucket, [{put, RefKey, <<>>},
-                                    {put, IdxKey, IdxVal}], WriteOpts)
+            eleveldb:write(Bucket, [{put, IdxKey, IdxVal}], WriteOpts)
```
larshesel: The comment above should probably say: only write idx

PR #1260, opened 2019-07-16
Upgrade cuttlefish to the latest version 2.2.0

codeadict:
Uses https://github.com/Kyorai/cuttlefish which is the version being published to hex.pm, they have been updating it quite a bit since the version we had and RabbitMQ folks are contributing to it as well.

Not sure if pinning the version would be best or just keep develop and do a rebar3 upgrade to get that latest ref. Thoughts @larshesel ?


larshesel:
Pinning to the version is preferable I think - makes things more explicit - I think we were tracking a the developmen branch was due to lack of a proper release tag.

larshesel:
Thanks - and really great to see cuttlefish being active again - I hope we can help out a bit!

PR #1261, opened 2019-07-17
Improve large queue initialization performance

larshesel:


apps/vmq_server/src/vmq_lvldb_store_sup.erl
```diff
@@ -27,13 +27,15 @@
...
 -define(TABLE, vmq_lvldb_store_buckets).
+-define(TBL_MSG_INIT, vmq_lvldb_init_msg_idx).
 
 %% ===================================================================
 %% API functions
 %% ===================================================================
 
 start_link() ->
     {ok, Pid} = supervisor:start_link({local, ?MODULE}, ?MODULE, []),
+    _ = ets:new(?TBL_MSG_INIT, [public, named_table, ordered_set]),
```
dergraf: add `{read_concurrency, true}`

PR #1266, opened 2019-07-23
Queue migration blocks until queue is in drained state

dergraf:


apps/vmq_server/src/vmq_queue.erl
```diff
@@ -195,11 +192,17 @@ save_sync_send_event(Queue, Event) ->
...
-info(Queue) ->
-    gen_fsm:sync_send_all_state_event(Queue, info, infinity).
+save_sync_send_all_state_event(Queue, Event) ->
+    case catch gen_fsm:sync_send_all_state_event(Queue, Event, infinity) of
+        {'EXIT', {normal, _}} ->
+            {error, noproc};
+        {'EXIT', {noproc, _}} ->
+            {error, noproc};
+        {'EXIT', Reason} ->
+            exit(Reason);
```
larshesel: What's the reason for removing these clauses? Why aren't they needed? Same comment for the ones in get_opts?

larshesel: aah, these were moved... the diff between + and - and red and green...

Issue #1246, opened 2019-07-09
Add `username` modifier for `auth_on_register` hooks


dergraf:

dergraf:
https://github.com/vernemq/vernemq/blob/master/apps/vmq_server/src/vmq_mqtt_fsm.erl#L641

larshesel:
Note, to make it available in plugins the modifier checker should be amended to allow it: 

https://github.com/vernemq/vernemq/blob/master/apps/vmq_plugin/src/vmq_plugin_util.erl#L99

PR #1268, opened 2019-07-23
Add `username` modifier for `auth_on_register` hooks

codeadict:
- Added the ability to modify the username from the `auth_on_register`
hook.
- Changed the `vmq_diversity` plugin to support this new modifier.
- Changed the `vmq_webhooks` plugin to support this new modifier.

### DIFF
--- a/apps/vmq_server/test/vmq_subscribe_SUITE.erl
+++ b/apps/vmq_server/test/vmq_subscribe_SUITE.erl
@@ -34,8 +34,7 @@ end_per_testcase(_, Config) ->
 all() ->
     [
      {group, mqttv4},
-     {group, mqttv5},
-     {group, try_private}
+     {group, mqttv5}
     ].
 
 groups() ->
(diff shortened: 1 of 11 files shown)


codeadict:
**Question:** Should `username` be added to the `reg_modifiers` type on https://github.com/vernemq/vernemq_dev/blob/master/src/auth_on_register_hook.erl and https://github.com/vernemq/vernemq_dev/blob/master/src/auth_on_register_m5_hook.erl?

apps/vmq_server/test/vmq_subscribe_SUITE.erl
```diff
@@ -34,8 +34,7 @@ end_per_testcase(_, Config) ->
 all() ->
     [
      {group, mqttv4},
-     {group, mqttv5},
-     {group, try_private}
```
larshesel: nice catch - forgot to remove that one

larshesel:
I think that would be perfect if it was added there as well

larshesel:
@dergraf can you check that this was what you had in mind when opening #1246 ?

larshesel:
closes #1246

PR #1274, opened 2019-07-29
add missing case of prop_val with the binary Default

gpolakow:
After the `username` modifier was added to `auth_on_register` hook, the username could not get through `prop_val` because no `prop_val` clause for binary was found.


larshesel:
Added you to the THANKS file: https://github.com/vernemq/vernemq/blob/master/THANKS#L32

Thanks again for noticing this and submitting the PR :heart:

PR #1283, opened 2019-08-07
Fix issue where errors would be logged when the /health endpoint was called (#1281).

### LINKED ISSUE #1281: Health request produces error in error.log
### Environment

- VerneMQ Version: 1.9.0 (installed from .deb)
- OS: Ubuntu 18.04 64bit
- VerneMQ configuration (vernemq.conf): default
- Cluster size/standalone: Standalone and/or cluster

### Expected behavior

No error produced

### Actual behaviour

Opening health URL http://localhost:8888/health shows Status OK in browser, but it produces following errors in error.log for each such a request:

`2019-08-06 10:07:18.296 [error] <0.698.0>@cowboy_rest:call:1576 CRASH REPORT Process <0.698.0> with 0 neighbours crashed with reason: no try clause matching #{bindings => #{},body_length => 0,cert => undefined,charset => undefined,has_body => false,has_sent_resp => true,headers => #{<<"accept">> => <<"text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3">>,<<"accept-encoding">> => <<"gzip, deflate, br">>,<<"accept-language">> => <<"sk-SK,sk;q=0.9,cs;q=0.8,en-US;q=0.7,en;q=0.6,de;q=0.5">>,<<"connection">> => <<"keep-alive">>,<<"dnt">> => <<"1">>,<<"host">> => <<"localhost:8888">>,...},...} in cowboy_rest:call/3 line 1576
2019-08-06 10:07:18.303 [error] <0.698.0>@cowboy_rest:call:1576 Cowboy stream 2 with ranch listener {{127,0,0,1},8888} and connection process <0.694.0> had its request process exit with reason: no try clause matching #{bindings => #{},body_length => 0,cert => undefined,charset => undefined,has_body => false,has_sent_resp => true,headers => #{<<"accept">> => <<"text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3">>,<<"accept-encoding">> => <<"gzip, deflate, br">>,<<"accept-language">> => <<"sk-SK,sk;q=0.9,cs;q=0.8,en-US;q=0.7,en;q=0.6,de;q=0.5">>,<<"connection">> => <<"keep-alive">>,<<"dnt">> => <<"1">>,<<"host">> => <<"localhost:8888">>,...},...} in cowboy_rest:call/3 line 1576`

Same behavior on several different machines. Direct install into the system as well as install in docker container.

larshesel:
Fixes #1281 

Perhaps we have this issue in other `cowboy_rest` endpoints as well - I'll investigate.

### DIFF
--- a/apps/vmq_server/src/vmq_health_http.erl
+++ b/apps/vmq_server/src/vmq_health_http.erl
@@ -17,26 +17,12 @@
 -behaviour(vmq_http_config).
 
 -export([routes/0]).
--export([init/2,
-         allowed_methods/2,
-         content_types_provided/2,
-         reply/2]).
+-export([init/2]).
 
 routes() ->
   [{"/health", ?MODULE, []}].
 
 init(Req, Opts) ->
-    {cowboy_rest, Req, Opts}.
-
-allowed_methods(Req, State) ->
-    {[<<"GET">>], Req, State}.
-
-content_types_provided(Req, State) ->
-	{[
-		{<<"application/json">>, reply}
-    ], Req, State}.
-
-reply(Req, _State) ->
     {Code, Payload} = case check_health_concerns() of
       [] ->
         {200, [{<<"status">>, <<"OK">>}]};
@@ -45,8 +31,8 @@ reply(Req, _State) ->
                {<<"reasons">>, Concerns}]}
     end,
     Headers = #{<<"content-type">> => <<"application/json">>},
-    cowboy_req:reply(Code, Headers, jsx:encode(Payload), Req).
-
+    cowboy_req:reply(Code, Headers, jsx:encode(Payload), Req),
+    {ok, Req, Opts}.
 
 -spec check_health_concerns() -> [] | [Concern :: string()].
 check_health_concerns() ->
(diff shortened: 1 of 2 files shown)


codeadict:
Ill check from the computer later, seems like the error happens due to the
content type, request was made with html content type and this responds to
json only.

larshesel:
> Ill check from the computer later, seems like the error happens due to the content type, request was made with html content type and this responds to json only.

I'm not so sure - the issue says that the body was correctly returned, but afterwards there's an error/crash logged.

codeadict:
Weirdly, the issue happens only when calling the endpoint from a browser. Neither "curl --header "Content-Type: application/json" http://localhost:8888/health" or "curl http://localhost:8888/health" cause the issue.

buzzdev:
Hi, thank you for looking into this. I came to this bug as i was reviewing the clustered setup. The loadbalancer checks the /health URL and i so many many of these errors caused by the probe, every 5 seconds.
So yes, i’m getting the issue by browser requests, but main concern is the LB probes, as the error logs would grow over time significantly.

Thank you

larshesel:
> `curl http://localhost:8888/health`

I can reproduce it every time with both the `curl --header "Content-Type: application/json" http://localhost:8888/health` and `curl http://localhost:8888/health`.

I think we'll merge this fix for now.

codeadict:
Sounds good

PR #1299, opened 2019-08-27
Add new `on_deliver*` hook variant and new `on_client_expired` hook

larshesel:
Old variants are deprecated and will be removed in the next major
version.

The new function signatures looks like follows:

```
on_deliver(User, SubscriberId, QoS, Topic, Payload, IsRetained)
on_deliver_m5(User, SubscriberId, QoS, Topic, Payload, IsRetain, Props)
```

Before merging this we should

- [ ] update webhooks or create issue to do so later
- [ ] update diversity or create issue to do so later
- [ ] update plugin hook docs or create issue to do so later
- [ ] update vernemq_dev or create issue to do so later

Created the issue #1346 to keep track of the above tasks


larshesel:
@dergraf I was wondering about the name `on_client_expired`. Perhaps a better name would be `on_session_expired`? The other two hooks are called `on_client_gone` and `on_client_offline` which refers to the actual client. The name `on_client_expired` doesn't make too much sense in that context I think.

PR #1310, opened 2019-09-05
Add new `max_last_will_delay` value and change default

larshesel:
The current default of `0s` was confusing to the users and as long as
it is possible from the broker side to limit this then it makes sense
for the default to be what the client provides. This can now be done
by specifying `client` instead of a time value.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -22,6 +22,7 @@ Mariano Guerra
 Szymon Kałasz
 Christoph Krey
 Michael Kuehne
+Will Lisac (@wlisac)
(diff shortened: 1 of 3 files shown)


wlisac:
@larshesel can this be merged or does it need to wait for a specific release?

larshesel:
I've postponed merging minor features as this makes it much easier to create bugfix releases. This will likely be merged either this or next week.

larshesel:
@wlisac merged! :)

PR #1356, opened 2019-10-31
Fix inflight offline message ordering

dergraf:
This PR addresses multiple issues, all discovered by fixing that the common tests use a 'real' message store. We haven't discovered this bug so far because our test suite disabled the message storage plugin by default (not sure why, performance?). Because of this queue compression never happened when running the vmq_server test suites.

The root cause of this issue is that the queue compresses queued messages when transitioning to offline state. During compression it loses information about message ids. This is an issue for messages with the DUP flag set, as we'd have to keep the message id to recover a qos1/qos2 flow.

### DIFF
--- a/apps/vmq_server/src/vmq_queue.erl
+++ b/apps/vmq_server/src/vmq_queue.erl
@@ -1025,27 +1025,32 @@ publish_last_will(#state{delayed_will = {_, Fun}} = State) ->
     Fun(),
     unset_will_timer(State#state{delayed_will = undefined}).
 
-maybe_offline_store(Offline, SubscriberId, #deliver{qos=QoS, msg=#vmq_msg{persisted=false} = Msg}=D) when QoS > 0 ->
-    PMsg = Msg#vmq_msg{persisted=true},
-    case vmq_plugin:only(msg_store_write, [SubscriberId, PMsg#vmq_msg{qos=QoS}]) of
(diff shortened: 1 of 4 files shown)


apps/vmq_server/src/vmq_queue.erl
```diff
@@ -1025,27 +1025,32 @@ publish_last_will(#state{delayed_will = {_, Fun}} = State) ->
...
+        %% a publish with the DUP flag set, and we must keep the
+        %% message id around (which isn't stored in the message store yet)
+        ok when Offline and IsDup ->
+            D#deliver{msg=PMsg};
         %% in case we have no online/serving session attached
         %% to this queue anymore we can save memory by only
         %% keeping the message ref in the queue
         ok when Offline ->
             PMsg#vmq_msg.msg_ref;
-        {ok, NewMsgRef} when Offline ->
```
dergraf: Forgotten dead code

apps/vmq_server/src/vmq_queue.erl
```diff
@@ -1025,27 +1025,32 @@ publish_last_will(#state{delayed_will = {_, Fun}} = State) ->
...
-            NewMsgRef;
         %% in case we still have online sessions attached
         %% to this queue, we keep the full message structure around
         ok ->
             D#deliver{msg=PMsg};
-        {ok, NewMsgRef} ->
```
dergraf: Forgotten dead code

PR #1372, opened 2019-11-19
Minor vmq_swc fixes and improvements

dergraf:
Refactorings:
- The references to Plumtree are deleted, as vmq_swc only uses the peer membership functionality from the origin Plumtree project, which aren't related to the actual Plumtree algorithm. We now use the 'auto' membership strategy instead of 'plumtree'.
- The transport module doesn't require the `swc_config` record for initiating a RPC, this enables that the RPC mechanism could be easier reused outside of `vmq_swc`
- The group membership server is now able to handle different 'membership_strategies', the default one is 'auto' (previously known as plumtree) . Alternatively it is now possible to set it to 'manual', enabling that an external process manages the peer membership.

Bug fixes:
- vmq_swc PUT operations didn't pass the old object context, they do now.

### DIFF
--- a/apps/vmq_swc/src/vmq_swc_store.erl
+++ b/apps/vmq_swc/src/vmq_swc_store.erl
@@ -140,7 +140,7 @@ lock(#swc_config{store=StoreName}) ->
     gen_server:call(StoreName, {lock, self()}, infinity).
 
 -spec remote_sync_missing(config(), peer(), [dot()]) -> [{db_key(), object()}].
-remote_sync_missing(#swc_config{group=SwcGroup, transport=TMod, peer=OriginPeer} = Config, RemotePeer, Dots) ->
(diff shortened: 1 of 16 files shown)


dergraf:
> Hard to understand what the problems were and what the fixes do. Can you add some more context?

See updated PR description above.

apps/vmq_swc/src/vmq_swc_store.erl
```diff
@@ -168,7 +168,7 @@ watermark(Config) ->
     get_watermark(Config).
 
 -spec remote_watermark(config(), peer()) -> watermark().
-remote_watermark(#swc_config{group=SwcGroup, transport=TMod} = Config, RemotePeer) ->
+remote_watermark(#swc_config{group=SwcGroup, transport=TMod}, RemotePeer) ->
```
larshesel: Where the `= Config` unused?

dergraf: Fixed

PR #1382, opened 2019-11-25
Upgrade cowboy dependency

dergraf:


dergraf:
We can't update atm as we run into a diaylzer issue, also reported here: https://github.com/ninenines/cowboy/issues/1378#issuecomment-558305215

Issue #1284, opened 2019-08-07
Beam.smp using 100% CPU


manish17mittal:
Issue-
We are trying to create an instant chat application with the machine specified as below. When verneMQ is run, initially the messages are transferring successfully between users. But within a small amount of time, the CPU spikes to 100%, and thereafter no further messages could be sent through the server. All the 3 machines of our cluster reach 100% CPU usage and our application stops working.

Environment-
- VerneMQ Version: 3.2.2
- OS: Ubuntu 18.04, 16 cores, 64 GB RAM
- Erlang/OTP version (if building from source): -
- VerneMQ configuration (vernemq.conf) or the changes from the default: only defined the IP address for the machine. No other change.
- Cluster size/standalone: Using cluster of 3 nodes
- Load balancer is used

ERROR LOGS================================================
1. HTOP-

PID USER      PRI  NI  VIRT   RES   SHR S CPU% MEM%   TIME+  Command                                                  13118 vernemq    20   0 16.3G 5239M  130M S 1600  8.3 31h01:04 
/usr/lib/vernemq/erts-10.2.3/bin/beam.smp -P 256000 -A 6413509 vernemq    20   0 16.3G 5239M  130M R 100.  8.3 55:07.46 
/usr/lib/vernemq/erts-10.2.3/bin/beam.smp -P 256000 -A 6413506 vernemq    20   0 16.3G 5239M  130M R 100.  8.3  1h25:01 
/usr/lib/vernemq/erts-10.2.3/bin/beam.smp -P 256000 -A 6413508 vernemq    20   0 16.3G 5239M  130M R 100.  8.3  1h06:38 
/usr/lib/vernemq/erts-10.2.3/bin/beam.smp -P 256000 -A 6413504 vernemq    20   0 16.3G 5239M  130M R 100.  8.3  1h38:14 
/usr/lib/vernemq/erts-10.2.3/bin/beam.smp -P 256000 -A 6413498 vernemq    20   0 16.3G 5239M  130M R 100.  8.3  2h24:16 
/usr/lib/vernemq/erts-10.2.3/bin/beam.smp -P 256000 -A 6413494 vernemq    20   0 16.3G 5239M  130M R 100.  8.3  2h45:47 
/usr/lib/vernemq/erts-10.2.3/bin/beam.smp -P 256000 -A 6413502 vernemq    20   0 16.3G 5239M  130M R 100.  8.3  1h50:21 
[...]

PR #1392, opened 2019-12-06
Upgrade plumtree to include fix for AAE crash

larshesel:


larshesel:
Should fix #1284 (at least the `LOCK` errors)

PR #1422, opened 2020-02-06
Add SwCGroup to Datadir path for RocksDB and Leveled backend

ioolkos:
The SWC plugin is pre-configured for 10 SWC groups,  backed by 10 LevelDB databases. To be compatible, this creates 10 DB directories for RocksDB and Leveled too.


ioolkos:
With this PR, we should be able to build a 'RocksDB' release for the SWC layer with:
`rebar3 as rocksdb release` (or `rebar3 as leveled release` to use Leveled)
Then configuring the SWC backend in the VerneMQ conf file:
`vmq_swc.db_backend = rocksdb`

This will still build and include LevelDB for the message store. Also, the RocksDB/Leveled backend is currently not user configurable (except maybe in an advanced.config file?). Also, it remains to be seen whether the 10 SWC groups is the right strategy here.

PR #1427, opened 2020-02-14
Expose cert chaing depth & Support multilevel bridge prefixes

ioolkos:
WIP. Adds:

- Configuration of cert depth for TLS bridges in the vernemq.conf file: `vmq_bridge.ssl.sbr0.depth =  4`
- Support for multilevel prefixes both for local and remote prefix, in configuring topics for bridges.

Note that you should not configure bridge topics with overlapping local subscriptions:
a/b/c and +/b/+ will both match for messages published to a/b/c.

TODO:
- [ ] tests
- [ ] check whether overlapping subscription cases work correctly

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # Changelog
 
 - Support multilevel bridge prefixes.
+- Make SSL cert depth configurable for bridges.
 
 ## VerneMQ 1.10.1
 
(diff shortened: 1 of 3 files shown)


changelog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+- Support multilevel bridge prefixes.
```
larshesel: Should the depth ssl cert thing be mentioned as well? Otherwise looks good to me.

Issue #1429, opened 2020-02-14
Connection error over ngxinx proxified websockets


cmnstmntmn:
### Environment

- VerneMQ Version: 1.10.1
- OS: Debian/Buster
- Erlang/OTP version (if building from source):
- Cluster size/standalone: single node

```yml
# vernemq.conf

allow_anonymous = on
listener.ws.default = 127.0.0.1:8080
listener.ws.proxy_protocol = on
```

```yml
#nginx proxy block

location /ws {
   proxy_pass http://127.0.0.1:8080/mqtt;
   proxy_http_version  1.1;
   proxy_cache_bypass  $http_upgrade;
   proxy_set_header Upgrade $http_upgrade;
   proxy_set_header Connection "upgrade";
   proxy_set_header Host $host;
   proxy_set_header X-Real-IP $remote_addr;
   proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
   proxy_set_header X-Forwarded-Proto $scheme;
   proxy_set_header X-Forwarded-Host $host;
   proxy_set_header X-Forwarded-Port $server_port;
}
```

### Expected behavior

Connect to VerneMQ via proxified ws listener.

### Actual behaviour

While trying to connect to **ws://url.tld/ws** i'm getting the following error on the verne side

```
2020-02-11 22:47:40.650 [error] <0.469.0> Ranch listener {{127,0,0,1},8080} terminated 
with reason: {{badmap,[{proxy_header,true}|#{env => #{dispatch => [{'_',[],[{[<<"mqtt">>],[],vmq_websocket,[{type,mqttws},{mountpoint,[]},
{allowed_protocol_versions,[3,4,131]},
{buffer_sizes,undefined}]}]}]},stream_handlers => [vmq_cowboy_websocket_h,cowboy_stream_h]}]},
[{maps,get,[proxy_header,[{proxy_header,true}|#{env => #{dispatch => [{'_',[],[{[<<"mqtt">>],
[],vmq_websocket,[{type,mqttws},{mountpoint,[]},{allowed_protocol_versions,[3,4,131]},{buffer_sizes,undefined}]}]}]},stream_handlers => [vmq_cowboy_websocket_h,...]}],...],...},...]}
```

@codeadict found that:

> So from the cowboy doc (the build-in webserver) the error indicates that you're proxy servers use the PROXY protocol:

```
[...]

PR #1430, opened 2020-02-16
fix proxy protocol header for websocket listeners

ioolkos:
Should address https://github.com/vernemq/vernemq/issues/1429


ioolkos:
Proxy protocol tests are failing, need to look into this.

Issue #1436, opened 2020-02-25
This command fails  vmq-admin session show --client_id=client1


orestes22:
### To help us save time and help you faster

vmq-admin session show --client_id=client1

This command fails when client_id parameter is a string composed by only numbers like "1234567890"

### Environment

- VerneMQ Version: 1.10
- OS: Kubernetes 
- Erlang/OTP version (if building from source):
- VerneMQ configuration (vernemq.conf) or the changes from the default
- Cluster size/standalone: 

### Expected behavior

- It should be work equal with number and characters 

### Actual behaviour

- This command fails when client_id parameter is a string composed by only numbers like "1234567890"

ioolkos:
Must be a bug, if it's not some issue about espacing characters in the shell.
Thanks for the observation @orestes22!

orestes22:
Thanks for read the issue. 
Also I want to add this behavior is the same using the API functionality.

Thanks and Best Regards

ioolkos:
@orestes22 can you test PR https://github.com/vernemq/vernemq/pull/1437 ?

orestes22:
Hi @ioolkos  , I dont have the correct environment to test it, right now Im using docket and kubernetes, so I only run and image in google cloud. 
Probably if you create a branch with this change i could test it. 

Best Regards

ioolkos:
Should be fixed by merged PR https://github.com/vernemq/vernemq/pull/1437

orestes22:
@ioolkos  Hi, good day, I recently try to use the last version, v1.10.1, this bug isnt fixed. 
Im not sure if you published this change or not,
Please, any feedback ill welcome. 
Best Regards

ioolkos:
No, it's not yet in a released package, but in master.
This stuff is easy to find out by the way...navigating around github.

orestes22:
Thanks for your answer, @ioolkos. I appreciate a lot. However, the main point of last question is when I will have the possibility to test it? 
Is there a schedule for the release?  Im pretty sure the team in charge will be very busy, but i need to implement an feature so i will love to know when i can update and start to program my feature. 
Best Regards

cmnstmntmn:
@orestes22 you may consider to build on your own from the master repo;
see [here](https://vernemq.com/downloads/index.html) how you can do it.

orestes22:
Hi, @cmnstmntmn Do you have experience in create a docker from the repo ? I was trying an im having some issue. Do you have any advice in this area ?
[...]

PR #1437, opened 2020-02-25
Fix number based client IDs in vmq-admin CLI. Fixes issue #1436.

ioolkos:


orestes22:
Hi, Good morning @ioolkos  @dergraf   When this will be ready ? Best Regards

Issue #1346, opened 2019-10-16
Update plugins and docs with new `on_deliver` and `on_session_expired` hooks


larshesel:
As mentioned in #1299 the new `on_deliver` and `on_session_expired` hooks should be supported in the plugins and properly documented:

- [ ] update webhooks
- [ ] update diversity
- [ ] update plugin hook
- [ ] update vernemq_dev

PR #1460, opened 2020-03-30
Update on_deliver* and add new on_session_expired hook  to vmq_webhooks and vmq_diversity

codeadict:
As mentioned in #1299 the new on_deliver and on_session_expired hooks should be supported in the plugins and properly documented. These changes fix https://github.com/vernemq/vernemq/issues/1346

- [x] update webhooks
- [x] update diversity
- [x] update plugin hook

Still have the documentation update left and this is using a branch of mine until the proposed change in `vernemq_dev` is released.


ioolkos:
Awesome, looks all good to me.
We probably need to tag this to a major release version, as it extends the plugin hooks API. Or maybe we can find a smart way to ensure backward compatibility for the extended hooks.

codeadict:
There is compatibility already with the old arity, the problem here lies when we release a new version that contains the new vernemq_dev it will break other plugins. One idea I got is to tag the current master as 1.0.0 and then the new release will be on par with the next VerneMQ version number. We definitely need good documentation and even a blog post to ensure people are informed enough.

ioolkos:
I guess we can keep the vernemq/vernemq_dev repo now, in the rebar.config file. We also need a short note for changelog.
Other than that, seems good to merge.

codeadict:
Perfect, I'll commit these changes.

ioolkos:
Base branch needs to be updated. Other than that, yeah, let's go forward and merge this into master.

codeadict:
Going to do it now

PR #1482, opened 2020-04-27
Add Observer CLI tool

ioolkos:


ioolkos:
Could also be interesting to add Observer CLI plugins to provide additional VerneMQ realtime information.

Issue #1477, opened 2020-04-24
vmq-admin plugin enable doesn't work as expected


cdro-idirect:
Hi all!

I'm trying to enable/disable a mqtt bridge without restarting VerneMQ. When looking at the documentation for the vmq-admin tool this seems to be achievable with just enabling/disabling the bridge plugin.

### Environment

- VerneMQ Version: 1.10.2
- OS: CentOS 8

### Behavior 
So, in my config I have configured a bridge, but I've set: plugins.vmq_bridge = off
Now I tried: vmq-admin plugin enable -n vmq_bridge
It displays "Done" and the bridge plugin is loaded, but the bridge that is configured is not connecting, so either I'm missing something here or there is a bug? Could someone please clarify?

(Actually it would be a nice feature for the vmq-admin to just have something like "vmq-admin bridge start/stop <name>" or something)

A little bit of context, in case you are wondering why I'm trying that: I'm trying to set up a cluster with multiple nodes that should connect via bridge to another broker/cluster. The thing is, I can only have one bridge active at a time, otherwise every node in the cluster will forward the message over it's own bridge, so I receive the same message multiple times in the other cluster.

ioolkos:
@cdro-idirect I think the plugin enable command should start any bridge that is pre-configured in the `vernemq.conf` file, yes. Do you see any errors logged?

I agree that having more options for bridge administration in `vmq-admin` would be nice. Certainly open to ideas and PRs.

cdro-idirect:
@ioolkos There is no error in the logfile. I changed the loglevel to debug and the only thing of interest that appears in the logfile is:
> 2020-04-27 06:28:01.076 [debug] <0.198.0>@vmq_plugin_mgr:load_app_modules:674 loading modules: [vmq_bridge,vmq_bridge_app,vmq_bridge_cli,vmq_bridge_sup]

`vmq-admin bridge show` gives no output:
> sudo vmq-admin bridge show
> Last login: Mon Apr 27 06:36:32 UTC 2020 on pts/1

The bridge config itself works. When I enable it in the config file and then restart vernemq, it automatically connects.

ioolkos:
@cdro-idirect ok, thanks. The "loading modules" debug log tells us that the bridge is actually correctly starting. 
So there seems to be some form of bootstrap issue in getting the bridge configs.

What happens if you do the following sequence: 
1. Start VerneMQ with `plugins.vmq_bridge = on`, 
2. Disable the plugin
[...]

PR #1490, opened 2020-04-29
Fix bridge plugin reloading

ioolkos:
This is hopefully a quick fix for https://github.com/vernemq/vernemq/issues/1477,
by adding a `vmq-admin bridge start` command that will start all bridges, after the bridge plugin has been enabled dynamically by using `vmq-admin plugin enable --name=vmq_bridge`


rebar.config
```diff
@@ -91,6 +92,7 @@
     {runtime_tools, load},
     {tools, load},
     {mcd, load},
+    {observer_cli, load},
```
codeadict, line 95: already included in the other PR or is this only for testing purposes?

Issue #1485, opened 2020-04-28
Customize Hostname Check and Wildcard SSL Certs


ioolkos:
There's an issue for outgoing SSL (as used in the VerneMQ bridge for instance).
When connecting to a server (another broker, in the bridge case) using a wildcard certificate, we'll fail with `{bad_cert, hostname_check_failed}`.
The reason is once more the strict TLS implementation of Erlang. The official recommendation is to add this to your SSL opts when doing the connection attempt:

`{customize_hostname_check,
           [{match_fun, public_key:pkix_verify_hostname_match_fun(https)}]}`

The problem with that is that we cannot add anonymous functions to config files! So, it seems we need to add that option at runtime, meaning it has to be added/implemented in the bridge.

codeadict:
Is this fixed now @ioolkos ?

ioolkos:
@codeadict thanks for reminder. I'm closing this as the config option for this has been merged. If it's a fix for the actual issue remains to be seen.

PR #1491, opened 2020-04-29
Fix wildcard cert bridge

ioolkos:
Potentially a fix for https://github.com/vernemq/vernemq/issues/1485 but currently untested.


rebar.config
```diff
@@ -12,6 +12,7 @@
 
 {deps, [
         {recon, "2.3.2"},
+        observer_cli,
```
codeadict, line 15: nice :)

PR #1498, opened 2020-05-07
README: add libsnappy requirement

rbino:
I was trying to compile VerneMQ 1.10.2 locally and it failed while compiling `eleveldb`.

I found the reason in [this comment](https://github.com/vernemq/eleveldb/pull/5#issuecomment-571895630), but since it's a little bit buried (and the error doesn't mention `libsnappy-dev`) I think it's better to add it to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ started, you need to first build VerneMQ.
 
 ### Building VerneMQ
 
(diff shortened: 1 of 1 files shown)


ioolkos:
thanks!

Issue #824, opened 2018-09-09
TCP connection timeout


XiaoKaiRen:
### Environment

- VerneMQ Version:
- OS: ubuntu 18.04
- Erlang/OTP version (if building from source):

My MQTT device will send the MQTT Connect Command to the Broker after 7 to 10 seconds of successful TCP connection because of the system limitation. But the Broker seems to have opened the connection before I have done this. How do I extend this timeout time?

larshesel:
Hi, thank you for opening the issue. If I understand it correctly you're running into the same issue as #735 which was only partially resolved and the PR therefore wasn't merged. Are you using TLS as well?

larshesel:
Would you be able to try out the code in https://github.com/erlio/vernemq/pull/736 and see if that solves your problem?

XiaoKaiRen:
Thankes this work!!
My problem is not from TSL
Just a small bug in the firmware.
then I just need to replace usr/lib/vernemq/lib/zmq_server-1.5.0/ebin/vmq_mqtt_fsm.beam  is it ok?

larshesel:
@dergraf do you think we should merge #736 even though it doesn't solve the issue for tls for the original issue #735?

@XiaoKaiRen I think that might work - but no guarantees - and I wouldn't recommend it as a long term solution ;)

PR #1527, opened 2020-06-02
Feature/set mqtt connect timeout

joskwanten:
This is a fix for #735 and #824 to make the mqtt connection timeout configurable. The timeout is the time between setting up the TCP or TLS connection and sending the 'CONNECT' message. The timeout time can be specified by setting the mqtt_connect_timeout in the vernemq.conf configuration file. The default is still 5000ms.

### DIFF
--- a/apps/vmq_server/priv/vmq_server.schema
+++ b/apps/vmq_server/priv/vmq_server.schema
@@ -333,7 +333,7 @@
                                                                          ]}.
 
 %% @doc specifies the connect timeout.
-{mapping, "connect_timeout", "vmq_server.connect_timeout", [
+{mapping, "mqtt_connect_timeout", "vmq_server.mqtt_connect_timeout", [
                                                             {default, 5000},
                                                             {datatype, integer},
                                                             hidden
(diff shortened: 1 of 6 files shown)


ioolkos:
@joskwanten thanks... did you actually encounter the mentioned issues, and this solves those?

joskwanten:
@ioolkos we did encounter this problem in the past when we tested an NB-IoT modem and typing AT commands by hand in a terminal to establish a connection to an MQTT server and were concerned that this can actually happen in a production environment on constrained devices in combination with a slow network. Therefore we want to be able to tune this timeout parameter if needed. Besides in other brokers like HiveMQ this is also a configurable setting (no-connect-packet-idle-timeout-millis). I tested this PR for both TCP and SSL and it will close the connection from server side when no CONNECT is send from the client in the specified time.

krisfremen:
@joskwanten recently encountered this issue as well. Thank you for the PR!

This would be a very welcome addition to VerneMQ!

ioolkos:
@joskwanten could you add an entry to the changelog? then we're good to go, I think.

codeadict:
This is awesome, tanks for it!

ioolkos:
Huh, travis checks still pending?

PR #1537, opened 2020-06-14
Catch ETS table call in vmq_cluster:netsplit_statistics/0

ioolkos:
Don't want to crash here when ETS table not yet available.
EDIT: this might mask but also unmask other issues. but the take here is that it's not worth crashing in just the netsplit stats call.


ioolkos:
WIP
- For a connecting client we want to terminate the client as we do not know whether the cluster is ready.
- For a vmq_systree call we do not want to terminate the caller
- For any internal calls in vmq_cluster coming from vmq_cluster_mon, we probably want to restart the cluster_mon, so that the vmq_status ETS table is re-created.

Issue #1326, opened 2019-09-20
Consider calling a callback on event of removing a subscription


manifest:
When a client that use `clean_session=true` have gone, all its subscriptions get removed. The issue that there is no way to find out about that action from within plugin.

It seems as both `on_unsubscribe` and `on_unsubscribe_m5` are only called on explicit unsubscribe message from the client.  The function call `vmq_subscriber_db:read(SubscriberId)` also return an empty list at the moment of `on_client_gone` execution.

The proposal is:
- to call `on_client_gone` callback before subscriptions of the client were removed, so that information will be available through the call to `vmq_subscriber_db:read(SubscriberId)`
- or call `on_unsubscribe` and `on_unsubscribe_m5` on any subscription removal (not only explicit unsubscribe message of the client)
- or add a new callback for that particular purpose

### Environment

- VerneMQ Version: 1.8.0
- OS: macOS, linux

larshesel:
Hi, thanks for the proposal. Can you describe the use case a bit more in detail - why is this feature necessary? 

> When a client that use `clean_session=true` have gone, all its subscriptions get removed. The issue that there is no way to find out about that action from within a plugin.

Changing the meaning of `on_subscribe/on_subscribe_m5` would not be an option as this would likely break existing plugins.

I don't think we cannot commit to option 1 as this would mean that we'd have to stick to the behaviour described forever, and the behaviour of `vmq_subscriber_db:read/1` is internal.

So I think the best option, should we decide to build support this feature, would be via option 3, unless we can think of some alternative.

manifest:
> Hi, thanks for the proposal. Can you describe the use case a bit more in detail - why is this feature necessary?

We're currently developing a platform where applications and end-users connected all together via MQTT. Some topics assigned to applications based on conventions. When an end-user subscibes to some topic of a particular application we interpret that as an intend of the end-user to consume application's resources.

In our case, that's essential to notify the application about the end-user's intend. We use on_subscribe callback of our VerneMQ plugin to send a message about the fact that subscription was created to the application. We also need a way to send a message about the fact that subscription was deleted, including the cases when the end-user closed its session or its session was closed by the broker, and that's where the issue rise.

In other words, we need a way to send a notification message when a subscription is created and when its deleted. The issue is that at the moment there is no way to find out about the subscription deletion from within a plugin.

manifest:
Just to generalize the description of the case and to clarify a reasoning behind proposed feature a little bit.

[...]

PR #1539, opened 2020-06-17
added on_topic_unsubscribed/1 hook

khodzha:
issue at hand: https://github.com/vernemq/vernemq/issues/1326

@ioolkos

### DIFF
--- a/apps/vmq_server/src/vmq_mqtt5_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt5_fsm.erl
@@ -581,7 +581,12 @@ connected(#mqtt5_unsubscribe{message_id=MessageId, topics=Topics, properties = P
     _ = vmq_metrics:incr(?MQTT5_UNSUBSCRIBE_RECEIVED),
     OnSuccess =
         fun(_SubscriberId, MaybeChangedTopics) ->
-                vmq_reg:unsubscribe(CAPSettings#cap_settings.allow_unsubscribe, SubscriberId, MaybeChangedTopics)
+                case vmq_reg:unsubscribe(CAPSettings#cap_settings.allow_unsubscribe, SubscriberId, MaybeChangedTopics) of
+                    ok ->
+                        vmq_plugin:all(on_topic_unsubscribed, [SubscriberId, MaybeChangedTopics]),
+                        ok;
+                    V -> V
+                end
         end,
     case unsubscribe(User, SubscriberId, Topics, Props0, OnSuccess) of
         {ok, Props1} ->
--- a/apps/vmq_server/src/vmq_mqtt_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt_fsm.erl
@@ -405,7 +405,12 @@ connected(#mqtt_unsubscribe{message_id=MessageId, topics=Topics}, State) ->
     _ = vmq_metrics:incr_mqtt_unsubscribe_received(),
     OnSuccess =
         fun(_SubscriberId, MaybeChangedTopics) ->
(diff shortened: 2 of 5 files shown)


ioolkos:
Thanks @khodzha.
Some things don't click for me. Can you confirm this implements your original issue? (it seemed to me it was about knowing which subscriptions were deleted in the plugin). I see though that this is more about *where* you want to call that hook, vs what the hook does.
We'll also need to think about the case where unsubscribe is not allowed due to CAP settings.

An alternative name would be `on_subscription_deleted` to move it away a little better from `on_unsubscribe`

khodzha:
yeah, you are right
i added topics parameter and `CAPSettings#cap_settings.allow_unsubscribe` check

codeadict:
This also will require tests and adding the hook to `vernemq_dev`.

khodzha:
vernemq_dev pr:
https://github.com/vernemq/vernemq_dev/pull/14
not sure about naming of `maybe_topics()` type

khodzha:
added tests

khodzha:
`vmq_rate_limiter` were failing so i tried to fix  them in https://github.com/vernemq/vernemq/pull/1544

khodzha:
@ioolkos any updates here? thanks

ioolkos:
@khodzha aware you're waiting... thanks for your patience in any case.
Quick question: you added MQTT5 tests but only implemented the hook for MQTT4 FSM? what if the queue is connected to an MQTT5 session?

khodzha:
@ioolkos thanks for catching this

the hook is already called from `vmq_reg:unsubscribe/3`
so i guess its not needed to call the hook explicitly in both FSMs, since there are these callback funs:
https://github.com/vernemq/vernemq/blob/947df6413e454adbf4cdbe5e9f7f92663d7858a3/apps/vmq_server/src/vmq_mqtt_fsm.erl#L406-L409
and
https://github.com/vernemq/vernemq/blob/947df6413e454adbf4cdbe5e9f7f92663d7858a3/apps/vmq_server/src/vmq_mqtt5_fsm.erl#L582-L585

ioolkos:
Hm, I still think it needs to be in the FSMs. What if you land here: 
https://github.com/vernemq/vernemq/blob/947df6413e454adbf4cdbe5e9f7f92663d7858a3/apps/vmq_server/src/vmq_mqtt_fsm.erl#L415

You might have set an ID to unsubscribed in your plugin/DB, being actually still subscribed on the broker. So, I think you want to wait for the success case, and then call the plugin. Or possibly check for successful feedback of vmq_reg function in OnSuccess fun, bet then it'd still be in the FSM.
Hope this makes sense.

khodzha:
@ioolkos sorry i maybe reading it wrong but i think `{error, _Reason}` clause is entered only when this branch is run and fails:
https://github.com/vernemq/vernemq/blob/947df6413e454adbf4cdbe5e9f7f92663d7858a3/apps/vmq_server/src/vmq_reg.erl#L104-L106

but `vmq_cluster:if_ready` shortcircuits without running `unsubscribe_op` if there is an error:
https://github.com/vernemq/vernemq/blob/947df6413e454adbf4cdbe5e9f7f92663d7858a3/apps/vmq_server/src/vmq_cluster.erl#L91-L97

khodzha:
rebased on latest master

khodzha:
@ioolkos

ioolkos:
@khodzha it'd be the only plugin we call in `vmq_reg`. This might be ok, but it just makes me a little cautious. In the sense of either we can do it with other plugins too, or there is a reason we keep everything in the MQTT FSMs.

khodzha:
i have no hard opinion on this, so why not
moved hook to FSMs

ioolkos:
@khodzha thanks, merged.

PR #1544, opened 2020-06-22
fixed rate limiter publish_rate_limit_test occasionally failing

khodzha:


ioolkos:
@khodzha Thanks, looks good to me. Only change outside test is that we put a TRef in init, right? (which then just sits there)

khodzha:
yeah, that's right; also take a closer look at my 10.5 secs in test, i think that should be a correct value but I'm not sure

ioolkos:
Let's see if this fixes tests on other PR.

PR #1564, opened 2020-07-11
WIP: Improve the memory queue buffering messages in the Bridge client

ioolkos:
- [x] ensure new incoming messages are added to buffer when buffer size is > 0 (to ensure message order). When buffer is deactivated, publish directly.
- [ ] add tests for above
- [x] ensure `vmq-admin bridge show` shows correct size and drop sizes for bridge at all times & add MQTT client mailbox size info to output
- [ ] add more detailed tests for buffer & test to ensure correct handling of waiting acks and packet ids.
- [x] expand the pluggable_metrics with counters on how many messages the bridge has consumed and published, separated by QoS
- [ ] New feature: add re-publisher to persistent queue (or similar feature) on bridge source broker, as an alternative to memory only buffering.
- [x] Handle Subscribe failure, by some form of re-subscribe strategy when encountering a return code 128 (fail to subscribe).
EDIT: Note: logging a warning for failed subs is good enough for now. `restart_timeout` in source broker should not be too low, so that chances are minimised we're hitting a starting up remote broker, can connect, but encountered failed subscriptions due to broker not fully initialised.
- [x] Allow multiple bridges between 2 same nodes, by identifying bridges by {Name, Host, Port} instead of {Host, Port} only.

### DIFF
--- a/apps/vmq_commons/src/gen_mqtt_client.erl
+++ b/apps/vmq_commons/src/gen_mqtt_client.erl
@@ -13,7 +13,7 @@
 %% limitations under the License.
 
 -module(gen_mqtt_client).
--behavior(gen_fsm).
+-behaviour(gen_fsm).
 -include("vmq_types.hrl").
 
 -ifdef(nowarn_gen_fsm).
@@ -291,8 +291,16 @@ connected({unsubscribe, Topics} = Msg, State=#state{transport={Transport, _}, so
                                   State#state{msgid='++'(MsgId),
                                               info_fun=NewInfoFun})};
 
+connected({publish, PubReq}, #state{o_queue=#queue{size=Size} = _Q} = State) when Size > 0 ->
+    NewState = maybe_queue_outgoing(PubReq, State),
+    {next_state, connected, NewState};
+                                            
 connected({publish, PubReq}, State) ->
     {next_state, connected, send_publish(PubReq, State)};
+                                            
+connected({publish_from_queue, PubReq}, State) ->
+    State1 = send_publish(PubReq, State),
+    {next_state, connected, maybe_publish_offline_msgs(State1)};
 
 connected({retry, Key}, State) ->
     {next_state, connected, handle_retry(Key, State)};
@@ -654,8 +662,8 @@ maybe_publish_offline_msgs(State) -> State.
 
 publish_from_queue(#queue{size=Size, queue=QQ} = Q, State0) when Size > 0 ->
     {{value, PubReq}, NewQQ} = queue:out(QQ),
-    State1 = send_publish(PubReq, State0),
-    maybe_publish_offline_msgs(State1#state{o_queue=Q#queue{size=Size-1, queue=NewQQ}}).
+    gen_fsm:send_event(self(), {publish_from_queue, PubReq}),
+    State0#state{o_queue=Q#queue{size=Size-1, queue=NewQQ}}.
 
 drop(#queue{drop=D}=Q) ->
     put(?PD_QDROP, D+1),
--- a/apps/vmq_commons/src/gen_mqtt_client.erl
+++ b/apps/vmq_commons/src/gen_mqtt_client.erl
@@ -547,10 +547,12 @@ handle_event(Event, StateName, State) ->
 
 handle_sync_event(info, _From, StateName,
                   #state{o_queue=#queue{size=Size,drop=Drop,max=Max}}=State) ->
+    {message_queue_len, Len} = erlang:process_info(self(), message_queue_len),
     Info =
         #{out_queue_size=>Size,
           out_queue_dropped=>Drop,
-          out_queue_max_size=>Max},
+          out_queue_max_size=>Max,
+          process_mailbox_size => Len},
     {reply, {ok, Info}, StateName, State};
 handle_sync_event(Req, From, StateName, State) ->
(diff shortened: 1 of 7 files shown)


ioolkos:
2 Push Bridges running from source broker to remote listener 10053. Added output for internal MQTT client mailbox (which should most of the time be 0).
```
➜  vernemq git:(fix_bridge_buffer) ✗ _build/default/rel/vernemq/bin/vmq-admin bridge show               
+-----------------+-------------+------------+---------------------+--------------------------+
| endpoint        | buffer size | buffer max | buffer dropped msgs | MQTT process mailbox len |
+-----------------+-------------+------------+---------------------+--------------------------+
| 127.0.0.1:10053 | 0           | 200000     | 0                   | 0                        |
+-----------------+-------------+------------+---------------------+--------------------------+
| 127.0.0.1:10053 | 0           | 200000     | 0                   | 0                        |
+-----------------+-------------+------------+---------------------+--------------------------+
```

ioolkos:
```
➜  vernemq git:(fix_bridge_buffer) ✗ _build/default/rel/vernemq/bin/vmq-admin bridge show
+------+-----------------+-------------+------------+---------------------+--------------------------+
| name | endpoint        | buffer size | buffer max | buffer dropped msgs | MQTT process mailbox len |
+------+-----------------+-------------+------------+---------------------+--------------------------+
| br1  | 127.0.0.1:10053 | 0           | 200000     | 0                   | 0                        |
+------+-----------------+-------------+------------+---------------------+--------------------------+
| br0  | 127.0.0.1:10053 | 0           | 200000     | 0                   | 0                        |
+------+-----------------+-------------+------------+---------------------+--------------------------+
```

ioolkos:
```
➜  vernemq git:(fix_bridge_buffer) ✗ _build/default/rel/vernemq/bin/vmq-admin metrics show |grep bridge
counter.vmq_bridge_publish_in_2 = 0
counter.vmq_bridge_dropped_msgs = 0
counter.vmq_bridge_publish_in_0 = 48843
counter.vmq_bridge_publish_out_2 = 0
counter.vmq_bridge_publish_out_0 = 0
counter.vmq_bridge_publish_in_1 = 0
counter.vmq_bridge_publish_out_1 = 0
```
Adding a couple metrics on what the bridge sends or receives. This is only what's received or pushed as MQTT packets (think on the wire).

For PUSH bridges, this will count `publish_out`. For PULL bridges this will count `publish_in`.

The internal subscription at the source broker is not counted in a PUSH bridge.
The internal publish at the source broker is not counted in a PULL bridge.

The new metrics above are unrelated to the size of the memory queue in a PUSH bridge.

codeadict:
So with this approach, if there is nothing in the buffer it will just publish directly instead of storing it right? Do you think a custom metric for the buffer size in bytes will be useful here? I personally like the idea of knowing how much memory the buffer is using.

apps/vmq_commons/src/gen_mqtt_client.erl
```diff
@@ -291,8 +294,16 @@ connected({unsubscribe, Topics} = Msg, State=#state{transport={Transport, _}, so
                                   State#state{msgid='++'(MsgId),
                                               info_fun=NewInfoFun})};
 
+connected({publish, PubReq}, #state{o_queue=#queue{size=Size} = _Q} = State) when Size > 0 ->
```
ioolkos, line 297: As long as the queue (buffer) is > 0, we keep buffering. When the queue is empty, we switch to direct sending. Note that this is in connected state. This will not switch back to buffering when we are connected but overloaded (then we just load mailbox of gen_mqtt). This buffer just protects disconnected bridges.
We could probably change this & just send everything through the buffer though.

ioolkos:
@codeadict added comment in place.
Re: memory insights regarding bridge buffer. probably makes sense, although easy to figure out looking at processes.
You have to configure a buffer max anyway. Normally taking your max message and multiply by buffer size, to have an idea on how to configure buffer max.

ioolkos:
Example for 2 configured bridges (br0, br1) on same source broker to same remote broker. br0 is a pull bridge, br1 is a push bridge, hence the different active counters.
```
➜  vernemq git:(fix_bridge_buffer) ✗ _build/default/rel/vernemq/bin/vmq-admin metrics show | grep bridge
counter.br1.vmq_bridge_publish_out_1 = 0
counter.vmq_bridge_dropped_msgs = 0
counter.br0.vmq_bridge_publish_in_0 = 17677
counter.br0.vmq_bridge_publish_in_1 = 0
counter.br1.vmq_bridge_publish_in_0 = 0
counter.br1.vmq_bridge_publish_out_2 = 0
counter.br0.vmq_bridge_publish_out_1 = 0
counter.br1.vmq_bridge_publish_in_1 = 0
counter.br0.vmq_bridge_publish_in_2 = 0
counter.br1.vmq_bridge_publish_in_2 = 0
counter.br0.vmq_bridge_publish_out_0 = 0
counter.br1.vmq_bridge_publish_out_0 = 22438
counter.br0.vmq_bridge_publish_out_2 = 0
```

ioolkos:
Going to merge as is, keeping 2 points open: async send implementation, and the persistence queue. 

I have implemented async send (with no suspension of port) and tested it, but I'm not sure what I can conclude. In any case, instead of suspension, we'd pretty much have to reconnect on a TCP level.

PR #1583, opened 2020-07-31
Minor edits in view of better Edoc use in the future.

ioolkos:


ioolkos:
Bare minimum to make `rebar3 edoc` pass.
The ultimate goal is to have better developer documentation and better specs/docs in modules.

PR #1630, opened 2020-09-15
Fix status page when behind proxy

seanwinterberger:
## Proposed Changes

Currently if running VMQ behind a proxy with say base-path being /vmq the status page fails to load.  The goal here is to request status.json relative to current path.  Currently we work around this using `sed` in our dockerfile but its a bit gross and this is such a small change figure open a PR for it.

## Types of Changes

## Checklist

### DIFF
--- a/apps/vmq_server/priv/static/js/status.js
+++ b/apps/vmq_server/priv/static/js/status.js
@@ -18,7 +18,7 @@ $(function() {
         template: $("#node_list_template").html(),
         target: $("#node_list"),
         cluster_status: {
-            url: "/status.json",
+            url: "status.json",
             last_calculated: Date.now(),
             rates: {}
         }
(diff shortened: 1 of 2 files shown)


ioolkos:
Thanks @seanwinterberger. I suppose you tested this works for the non-proxy case too. Seems good to me.

seanwinterberger:
> 
> 
> Thanks @seanwinterberger. I suppose you tested this works for the non-proxy case too. Seems good to me.

Yah, I do not run a proxy on the cluster locally but I will verify in a bit here.

ioolkos:
@seanwinterberger thanks! can you also add a note (something like the commit message is fine) to the changelog?

seanwinterberger:
Yah I will add a comment in the changlog, here is example of it working in both scenarios.
image
image

ioolkos:
@seanwinterberger jep, tested it too, it works. Good to merge after changelog commit.

ioolkos (approved):

PR #1646, opened 2020-10-02
replaced all_topics atom in on_topic_unsubscribed hook with actual subscriptions list

khodzha:
## Proposed Changes

Right now `on_topic_unsubscribed` hook either passes a list of topics or an atom `all_topics`, ie the signature is
`on_topic_unsubscribed(SubscriberId    :: subscriber_id(), MaybeTopics     :: all_topics | [topic()])`

1) `all_topics` is kind of lazy
2) An assumption that you can query these topics from `vmq_subscriber_db` is implicit. This non-obvious feature could be broken while moving code around, since topics aren't explicitly passed with the hook.

This should improve this hook API by unifying both cases into
`on_topic_unsubscribed(SubscriberId    :: subscriber_id(), Topics :: [topic()])`
which looks clearer in my opinion.

I marked this is as breaking change since technically it probably is.
But I think it won't actually break something as of now because:
1) the left case is probably covering removed `all_topics` case fine
2) the hook is very new and i suppose isn't widespread

## Types of Changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Cosmetics (whitespace, styles...)

## Checklist

### DIFF
--- a/apps/vmq_server/src/vmq_queue.erl
+++ b/apps/vmq_server/src/vmq_queue.erl
@@ -438,7 +438,8 @@ offline({enqueue, Enq}, #state{id=SId} = State) ->
     {next_state, offline, insert(Enq, State)};
 offline(expire_session, #state{id=SId, offline=#queue{queue=Q}} = State) ->
     %% session has expired cleanup and go down
-    vmq_plugin:all(on_topic_unsubscribed, [SId, all_topics]),
+    UnsubbedTopics = get_topics(SId),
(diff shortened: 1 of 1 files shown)


ioolkos:
@khodzha I suppose it makes sense. THB have never had a use for `on_topic_unsubscribed` so far.
The only consideration here is that we're hitting the subscriber store for every subscriber, in "massive situations", like massive forceful disconnects etc.

PR #1653, opened 2020-10-10
Feature LMDB messagestore

ioolkos:
Want to check how LMDB could work as a backend for offline queues (or offline messages, to be more specific).
This is just a quick draft.
- [ ] adapt message store tests
- [ ] do some manual basic function testing
- [ ] make production version of the engine
- [ ] add a metadata (SWC) backend with LMDB too.

This initiative was started by @codeadict and uses the https://github.com/codeadict/elmdb NIF which already has a couple fixes.


ioolkos:
whoops... branched off of feature_rocksdb_messagestore.

Issue #1560, opened 2020-07-02
Ensure that MQTT v5 user properties are stored to disk/DB with a message


ioolkos:
entry point: 
https://github.com/vernemq/vernemq/blob/379c7d15871ede362ae19dd4a35621ba5928b024/apps/vmq_generic_msg_store/src/vmq_generic_msg_store.erl#L284

PR #1655, opened 2020-10-12
Ensure_that_MQTT_v5_user_properties_are_stored_to_disk/DB_#1560

steve999lee:
…age_#1560

## Proposed Changes

Changes to vmq_generic_msg_store_utils.erl to read and write properties from the store. Payload and Properties are combined into a single tuple before writing. 

The change is backwards compatible in case the store contains messages persisted by an earlier version 

I modified an existing test so that test messages contain properties

## Types of Changes

- [x ] Bugfix (non-breaking change which fixes issue #1560 )

## Checklist

- [x ] I have read the `CODE_OF_CONDUCT.md` document
- [x ] All tests pass locally with my changes
- [x ] I have added tests that prove my fix is effective or that my feature works
- [x ] I have squashed all my commits into one before merging

## Further Comments

This is my first attempt at erlang code so probably the style isnt great

### DIFF
--- a/apps/vmq_generic_msg_store/src/vmq_generic_msg_store.erl
+++ b/apps/vmq_generic_msg_store/src/vmq_generic_msg_store.erl
@@ -283,7 +283,7 @@ code_change(_OldVsn, State, _Extra) ->
 %%%===================================================================
 handle_req({write, {MP, _} = SubscriberId,
             #vmq_msg{msg_ref=MsgRef, mountpoint=MP, dup=Dup, qos=QoS,
-                     routing_key=RoutingKey, payload=Payload}},
+                     routing_key=RoutingKey, properties=Properties, payload=Payload}},
            #state{engine=EngineState, engine_module=EngineModule, refs=Refs}) ->
     MsgKey = sext:encode({msg, MsgRef, {MP, ''}}),
     IdxKey = sext:encode({idx, SubscriberId, MsgRef}),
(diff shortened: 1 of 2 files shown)


ioolkos:
@steve999lee thanks! will review asap. The build currently fails due to type errors. (if you want to check & fix this locally, run `./rebar3 dialyzer`).
Protocol backward compatibility is vital, obviously, but you seem to have checked for that. If this proves good enough, it'll be a great contribution! :)

ioolkos:
Locally, I have 2 errors in the `vmq_publish SUITE`,
one for `mqttv4.pattern_matching_test`, one for `mqttv4.not_allowed_publish_close_qos2_mqtt_3_1_1`, so I need to investigate those.

steve999lee:
Ah - I only ran the ct tests - I guess I have to run dialyzer and eunit as well

ioolkos:
@steve999lee `./rebar3 ct` will run all the tests. `./rebar3 dialyzer` has to be run separately.
Apparently, for you all the Common Tests pass?

steve999lee:
@ioolkos - I think so

Skipped 14 (14, 0) tests. Passed 389 tests.

steve999lee:
I can fix it such that dialyzer and ct (and my tests) pass by changing the type to allow either a payload() or a tuple() 
`-type p_msg_val_pre() :: {routing_key(), payload() | tuple()}.`

Or perhaps there is a better way?

ioolkos (approved):

ioolkos:
@steve999lee thanks again for your effort, I hope this simple approach can take us far. If you test this on MQTT v5 level, can you observe behaviour and semantics a little bit? (like: are message TTL respected, do all properties arrive at subscriber, do we potentially need to filter out properties at some point, etc.)

steve999lee:
@ioolkos - yes we plan to use MQTT5 for our messaging framework and to make use of properties and TTL so we are thoroughly testing these behaviours. I don't have a use case for filtering out properties though.

PR #1666, opened 2020-11-02
Attempt to allow inet6 in vmq-admin calls

ioolkos:


ioolkos:
Beware, this is explorative and does not work.

Issue #1703, opened 2021-01-13
Websocket listener dont process proxy_protocol option


mtaobiz:
### Environment

- VerneMQ Version: 1.11.0
- OS: Debian 10.4
- Erlang/OTP version (if building from source):
- VerneMQ configuration (vernemq.conf) or the changes from the default

listener.tcp.default = 127.0.0.1:1883
listener.tcp.listener1 = 10.1.4.180:1883
listener.tcp.proxy1 = 10.1.4.180:1884
listener.tcp.proxy1.proxy_protocol = on

listener.ws.default = 0.0.0.0:9002
listener.ws.default.proxy_protocol = on

- Cluster size/standalone:

clustered

### Expected behaviour

Websocket connecttion from browser through nginx and haproxy accepted by vernemq

### Actual behaviour

Error during WebSocket handshake: Unexpected response code: 400

proxy_protocol option ignored in ws listener or don`t work

- Please include minimal steps to reproduce the issue, if possible.

nginx conf:
```
...
upstream ws_proxy {
    least_conn;
    server 127.0.0.1:9002;
}

server {
listen 443 ssl;
...
location /ws_proxy {
    proxy_pass http://ws_proxy/mqtt;
    proxy_redirect          default;
    proxy_pass_request_headers            off;
    proxy_http_version                    1.1;
    proxy_set_header   Host               $host;
    proxy_set_header   Connection         "Upgrade";
    proxy_set_header   Upgrade            $http_upgrade;
    proxy_set_header   User-Agent         $http_user_agent;
    proxy_set_header   Origin             $http_origin;
    proxy_set_header   Sec-WebSocket-Version $http_sec_websocket_version;
    proxy_set_header   Accept-Encoding    $http_accept_encoding;
    proxy_set_header   Accept-Language    $http_accept_language;
    proxy_set_header   Sec-WebSocket-Key  $http_sec_websocket_key;
    proxy_set_header   Sec-WebSocket-Extensions  $http_sec_websocket_extensions;
    proxy_set_header   Sec-WebSocket-Protocol  $http_sec_websocket_protocol;
    proxy_set_header   X-Real-IP          $remote_addr;
    proxy_set_header   X-Forwarded-For    $remote_addr;
    proxy_set_header   X-Forwarded-Proto  $scheme;
}
}
```
haproxy.conf:
```
defaults                            
        log     global
        mode    http
        option  httplog
        option  dontlognull
        timeout connect 5000        
        timeout client  50000       
[...]

PR #1739, opened 2021-02-23
fix ws options in proxy_protocol case

ioolkos:
- [x] Add test, ensuring proxied IP/Port works
- [ ] Add test for SSL and common name as username


ioolkos:
Fixes https://github.com/vernemq/vernemq/issues/1703

Issue #1700, opened 2021-01-06
VernemQ 3-node Cluster High-Availability : Message loss on node fault with QoS1


MarcVanOevelen:
### Related issues :

 - Message loss on broker fault with QoS 1 with retained session (#1577)
 - High-Availability Support and Assumptions (#1628)

#### DUT

VerneMQ 3-node cluster deployment in k8s :

 - kubernetes : v1.17.6 deployment in 5 VMs (1 master and 4 minions)
 - deployment : using the provided helm chart
 - vernemq image : vernemq/vernemq:1.11.0-alpine
 - configuration changes from the default helm config :
 ```
      replicaCount: 3
      persistentVolume: enabled: true
      service:
        type: NodePort
        mqtt:
          nodePort: 30883 
```
With this configuration the cluster mqtt port is accessible at the k8s ip_address on port 30883
The k8s service "loadbalances" external mqtt connections over the 3 cluster nodes.

##### Clients

2 clients running on a remote workstation connect to the cluster :

 - Txm: "transmitter" or "publisher"
   - MQTT 3.1.1
   - qos = 1
   - clean_session = false
   - retain = false
 - Rxm: "receiver" or "subscriber"
   - MQTT 3.1.1
   - qos = 1
   - clean_session = false
   - retain = false

##### Message definition :

  - key = numerical index formatted as string
  - value = a json structure containing :

key | example value | description
----- | ------------------- | ---------------
tx-id |	"txm-0" | string identifying the Txm instance
txproc-id |	"proc_0" | string identifying the tx process in the Txm instance
tx-ix |	24 | transmit index, same value as the key
tx-eos | false | flag indicating the end-of-sequence
tx-timestamp | 1607951854126872375 | timestamp in nanoseconds since epoch
keyspace-size | 32 | keyspace size
payload	| "" | optional extra payload bytes

##### Publishing :

Txm publishes a single-shot keyspace consisting of 1000 messages with tx-ix ranging from 0 to 999
The sequence of message values is published to topics "testdata/\<tx-id\>/\<tx-ix\>"
i.e.   "testdata/txm-0/0" 
       "testdata/txm-0/1" 
       ...
       "testdata/txm-0/999" 

Publish rate is set to about 50 messages/sec

##### Subscription :

Rxm subscribes to the Txm keyspace and consumes the messages :
  - computes latency between the embedded tx-timestamp and the current rx-timestamp
  - analyses the received indexes for "missing" and "duplicate" messages
[...]

PR #1769, opened 2021-04-07
Cluster ha improvements

MarcVanOevelen:
## Proposed Changes

These changes fix missing messages in QoS1 in a HA VerneMQ cluster
as reported in this issue:
Message loss on node fault with QoS1(#1700)

Following problems were discovered/addressed:

1. Make enqueue a synhronous call :
When the broker receives a message from a publisher it was written into the subscriber(s) queue(s)
using an asynchronous function call. This causes the ack to be sent before the message was written to persistent storage
hence leaving a small time-window in which a crash of the broker causes the acknowledged message to be lost.

Solved by using a synchronous call.

2. Add handshake on inter-node TCP connection :
When the broker receives messages for a subscriber connected to another node, the messages are sent via the dedicated inter-node TCP connection.  A crash of the receiving node while a message is present in the tcp receive buffer
and acknowledged to the sender but not yet processed by the broker will cause that message to be lost.
The publisher received an ack when the message was delivered to the senders tcp stack.

Solved by introducing an application-level handshake on the inter-node tcp connection:
 - The message block "in transit" is buffered until acknowledged by the receiving node.
 - In case the receiving node crashed and hence no ack was received, the buffer is resent when the connection is re-established.
 - In case the subscriber reconnected to another node, the resent messages are forwarded to that new node.

3. Catch crashes during queue migration :
During HA testing some some crashes were observed when queue migration was triggered
i.,e. rpc call to get queue pid from the crashed node and ets:lookup.

Solved by handling the exceptions.

4. Only release messages when acknowledged :
When the broker delivers messages to a connected subscriber, the messages from the queue are saved in a
backup queue and "mailed" to the fsm process that performs the interaction with the client. (retries, acks, ...).
But the backup messages were released (=deleted from persistent storage) before they were actually acknowledged
by the subscriber.
Again, when the broker crashes those messages are lost.

Solved by : 
- releasing each message individually when the fsm receives the ack.
- QoS0 messages are released immediately (unless potentially being upgraded by "maybe_upgrade_qos")
- next message batch is now added to the backup queue instead of replacing it.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #1700)

## Checklist

## Further Comments

These fixes do not yet cover all cases of message loss but already drastically reduced their ocurrence:

Without the fixes about 1 in 3 HA testruns showed message loss.
Above fixes reduced this to about 1 lost message in 1000 testruns.

### DIFF
--- a/apps/vmq_server/src/vmq_cluster_node.erl
+++ b/apps/vmq_server/src/vmq_cluster_node.erl
@@ -37,14 +37,18 @@
           socket,
           transport,
           reachable=false,
+          waiting_for_ack=false,
           pending = [],
+          transit = [],
           max_queue_size,
           reconnect_tref,
+          waiting_ack_tref,
           async_connect_pid,
           bytes_dropped={os:timestamp(), 0},
           bytes_send={os:timestamp(), 0}}).
 
 -define(RECONNECT, 1000).
+-define(ACK_TIMEOUT, 3000).
 
 %%%===================================================================
 %%% API
@@ -120,9 +124,9 @@ init([Parent, RemoteNode]) ->
     erlang:send_after(1000, self(), reconnect),
     loop(#state{parent=Parent, node=RemoteNode, max_queue_size=MaxQueueSize}).
 
-loop(#state{pending=Pending, reachable=Reachable} = State)
+loop(#state{waiting_for_ack=Waiting, reachable=Reachable} = State)
   when
-      Pending == [];
+      Waiting == true;
       Reachable == false ->
     receive
         M ->
@@ -201,20 +205,32 @@ handle_message({connect_async_done, AsyncPid, {ok, {Transport, Socket}}},
     Msg = [<<"vmq-connect">>, <<L:32, NodeName/binary>>],
     case send(Transport, Socket, Msg) of
         ok ->
-            lager:info("successfully connected to cluster node ~p", [RemoteNode]),
-            State#state{socket=Socket, transport=Transport,
+            lager:info("cluster_node ~p : successfully connected to cluster node ~p", [self(), RemoteNode]),
+            NewState = State#state{socket=Socket, transport=Transport,
                         %% !!! remote node is reachable
                         async_connect_pid=undefined,
-                        reachable=true};
+                        reachable=true,
+                        waiting_for_ack=false},
+            internal_flush(NewState);
         {error, Reason} ->
-            lager:warning("can't initiate connect to cluster node ~p due to ~p", [RemoteNode, Reason]),
+            lager:warning("cluster_node ~p : can't initiate connect to cluster node ~p due to ~p", [self(), RemoteNode, Reason]),
             close_reconnect(State)
     end;
 handle_message({connect_async_done, AsyncPid, error}, #state{async_connect_pid=AsyncPid} = State) ->
     % connect_async already logged the error details
     close_reconnect(State);
+handle_message(block_ack, State) ->
+     %% block ack received, cancel timeout timer
+     NewState = cancel_ack_timer(State),
+     NewState#state{waiting_for_ack=false, transit=[]};
 handle_message(reconnect, #state{reachable=false} = State) ->
-    connect(State#state{reconnect_tref=undefined});
+    %% avoid timer interference
+    NewState = cancel_ack_timer(State),
+    connect(NewState#state{reconnect_tref=undefined});
+handle_message(reconnect, #state{reachable=true} = State) ->
+    %% avoid timer interference
+    NewState = cancel_ack_timer(State),
+    close_reconnect(NewState);
 handle_message({status, CallerPid, Ref}, #state{socket=Socket, reachable=Reachable}=State) ->
     Status =
     case Reachable of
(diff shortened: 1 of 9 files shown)


ioolkos:
@MarcVanOevelen thanks... what a great effort! :)
As it's a rather big change it'll take me a while to review. We also need to make sure that any "change of strategy" (from async to sync processing, for instance) has no downstream effects.

If you feel I need any more specific information to for an effective review, let me know. (I don't think so, for now at least).

MarcVanOevelen:
@ioolkos Thanks!

It took me quite some time to get familiar with the codebase but I must say it is very well structured,
and this is such a great project that it was really worth the effort!

Indeed such changes are very tricky given the many features and use-cases.
Hopefully they don't cause any side-effects that are not covered by the tests.

Our use case is focused on a small number of publishers/subscribers but we really need
reliable message delivery (QoS1) in a HA cluster.

I still have to assess the inevitable performance cost of those changes.

The release smoketest fails now but I suppose this is caused by the rebar3_cuttlefish dependency
which points to master. I had to specify the previous version to get this working again.

I also see that a test is failing which did not happen when I ran them locally?

If you need more info please let me know.
Thanks.

apps/vmq_server/src/vmq_cluster_node.erl
```diff
@@ -201,20 +205,32 @@ handle_message({connect_async_done, AsyncPid, {ok, {Transport, Socket}}},
...
+            NewState = State#state{socket=Socket, transport=Transport,
                         %% !!! remote node is reachable
                         async_connect_pid=undefined,
-                        reachable=true};
+                        reachable=true,
+                        waiting_for_ack=false},
+            internal_flush(NewState);
         {error, Reason} ->
-            lager:warning("can't initiate connect to cluster node ~p due to ~p", [RemoteNode, Reason]),
+            lager:warning("cluster_node ~p : can't initiate connect to cluster node ~p due to ~p", [self(), RemoteNode, Reason]),
```
codeadict, line 216: Should we use `self()` (the pid) here or `node()`?

apps/vmq_server/src/vmq_cluster_node.erl
```diff
@@ -256,12 +272,29 @@ maybe_flush(#state{pending=Pending} = State) ->
...
+%%  - append a 'tail' to the block of messages that are sent
+%%  - keep a backup copy of the messages in a 'transit' buffer
+%%  - await the block_ack from the receiving node, as it handled the complete block
+%%    (upon receipt of the 'tail')
+%%  - when the block_ack is received, the transit buffer is cleared
+%%    and the next block can be transmitted
+%% This handshake is protected by a timeout which will trigger a reconnect
+%% Note that currently the backup is kept only in memory, a future improvement
+%% would be to persist the backup on disk to also protect against message loss
+%% in case the transmitting node crashes.
```
codeadict, line 287: Maybe we should put this in a ticket @ioolkos so it doesn't get lost in the code comments

apps/vmq_server/src/vmq_mqtt5_fsm.erl
```diff
@@ -1420,6 +1421,22 @@ prepare_frame(#deliver{qos=QoS, msg_id=MsgId, msg=Msg}, State0) ->
                                      Msg#vmq_msg{qos=NewQoS}, WAcks)}}
     end.
 
+
```
codeadict, line 1424: Remove extra line here:
```suggestion
```

apps/vmq_server/src/vmq_mqtt5_fsm.erl
```diff
@@ -1420,6 +1421,22 @@ prepare_frame(#deliver{qos=QoS, msg_id=MsgId, msg=Msg}, State0) ->
...
+-spec release_message(msg(), state()) -> state().
+release_message(Msg, #state{queue_pid=QPid} = State) -> 
+     vmq_queue:release_message(QPid, Msg),
+     State.
+
+
+
+
+
```
codeadict, line 1439: And here...
```suggestion
```

apps/vmq_server/test/vmq_cluster_com_SUITE.erl
```diff
@@ -47,19 +50,49 @@ all() ->
 connect_params(_RemoteNode) ->
     {gen_tcp, {127,0,0,1}, 12345}.
 
-connect_success_test(Config) ->
+
+
+
```
codeadict, line 55:
```suggestion
```

apps/vmq_server/test/vmq_cluster_com_SUITE.erl
```diff
@@ -163,3 +244,44 @@ recv_message(Socket, Term) ->
...
+            io:format(user, "got ~p instead of ~p~n", [E, {ok, BatchMsg}]),
+            E
+    end;
+recv_message_block(Socket, ClusterNodePid, Msg) ->
+    recv_message_block(Socket, ClusterNodePid, [Msg]).
+
+ack_message_block(ClusterNodePid) ->
+    ClusterNodePid ! block_ack,
+    ok.
+
```
codeadict, line 271:
```suggestion
```

apps/vmq_server/test/vmq_queue_SUITE.erl
```diff
@@ -354,6 +355,12 @@ mock_session(Parent) ->
             ok
     end.
 
+release_msgs(_, []) -> ok;
+release_msgs(QPid, [Msg | Rest]) ->
+    vmq_queue:release_message(QPid, Msg),
+    release_msgs(QPid, Rest).
+
+
```
codeadict, line 363:
```suggestion
```

apps/vmq_server/src/vmq_cluster_node.erl
MarcVanOevelen: I agree, node() is better, will adapt it.

codeadict: Thanks 🎉

ioolkos:
Common Test suites and Dialyzer pass for me locally.

ioolkos:
@MarcVanOevelen no in-depth review yet, but I like your code/commenting style, very much aligned with existing code.

It'd be great to have an idea on performance impact, as you already mentioned. We also need to protect any RAM-grabbing component (buffer/queue/backup) or at least ensure that it will hit an upstream config limit (haven't yet looked into he details of that point though).

ioolkos:
@MarcVanOevelen I'll try to run some basic performance comparison over the weekend. Anything specific you'd want me to test?

MarcVanOevelen:
> @MarcVanOevelen I'll try to run some basic performance comparison over the weekend. Anything specific you'd want me to test?

I have successfully completed the HA behavior in our production system.
Now I am also running performance tests relative to vernemq:1.11.0-alpine deployment using helm chart in k8s.
These tests try to assess max sustainable message rates and latency using 1 publisher and 1 subscriber with following parameters:
  - QoS:  [0, 1, 2]
  - message sizes: [100 bytes, 64K bytes]
  - setup : [single node, 2-node cluster]  
       clients connect via k8s service, hence in 2-node cluster publisher and subscriber connect to different nodes
I almost completed testing but unfortunately we just suffered from a power outage.
Results will be ready in the course of next week.

Since performance tests are very dependent on the test-environment (resources, ...)  we can learn most from relative tests.
But it would be nice if you could test using the same parameters so we can compare our results.

ioolkos:
@MarcVanOevelen great, thanks. Hopefully no power outages any more :)

ioolkos:
Did a basic test to get a feeling for latencies. This is a quick test with 50 publisher on node 1 to 1 subscriber on node 2, sending 1500 100 byte messages total per second (QoS 1).
Highest line is max latency, then 95 percentile. Max line is at 90ms vs 30ms. So, this adds 60ms. 

1769 Branch:
1769 Branch
Main Branch
Master Branch

ioolkos:
@MarcVanOevelen the loop in `vmq_cluster_node` is racing through `loop(internal_flush)` because you've disabled the check for Pending in the first clause.

MarcVanOevelen:
> @MarcVanOevelen the loop in `vmq_cluster_node` is racing through `loop(internal_flush)` because you've disabled the check for Pending in the first clause.

Nice catch!
 I included the original "Pending" check again.
Thanks!

MarcVanOevelen:
@ioolkos performance test comparison results in attached pdf
[VerneMQ_performance_test.pdf](https://github.com/vernemq/vernemq/files/6350407/VerneMQ_performance_test.pdf)

ioolkos:
@MarcVanOevelen thanks... can you leave me your e-mail address at contact email here: https://vernemq.com/community.html

MarcVanOevelen:
@ioolkos  and here the HA test results
[VerneMQ_cluster_HA_test.pdf](https://github.com/vernemq/vernemq/files/6350606/VerneMQ_cluster_HA_test.pdf)

PR #1778, opened 2021-04-13
SWC iteration 4

ioolkos:
## Proposed Changes as DRAFT
SWC Iteration 4 will be part of the next VerneMQ release.

This specifically addresses the requirement to allow a node leave the cluster and re-join with the same nodename but empty state. (the latter causing K8s issues when Pods get killed, and a leave strategy is chosen instead of stop/restart node)

TODO:
- [x] clean from development logs/comments
- [x] Unit tests are switched to SWC but might need more adjustments, mainly due to different timing behaviour in sync protocol
- [x] Review peer ORSWOT
- [ ] test against every case determined or conjectured to be related to fix
EDIT: testing against the non-deterministic cases only partially done/possible.

Technical note on SWC Iteration 4:
Causality tracking needs Ids to track events under. SWC does causality tracking based on the different cluster nodes. Before, the nodename was used as the SWC id (example `'dev0@127.0.0.1'`). This is now `{'dev0@127.0.0.1', <<"some binary actor">>}`.
As an actor is created for an empty starting node, a node re-joining a cluster under the same nodename will now have a different SWC id. This improves causality tracking, but note that the old IDs are now kept in the nodeclocks.

SWC Iteration 5 should add jumping (cleaning) the clocks, and retiring Ids with transfer to a new ID.


mitchellrj:
@ioolkos am I reading this right, when I say the change to vars.config changes the default metadata plugin to SWC from plumtree?

If so, this is a breaking change that was not covered in the release notes and should be amended.

PR #1806, opened 2021-06-03
Fix boot script/Rebar3

ioolkos:
This updates `Rebar3` and copies start.boot to vernemq.boot, as required by old runner script.

@varnerac this should allow us to go with the latest `Rebar3`


varnerac:
Looks good. It passes the release smoke test, so all is well.

Issue #1780, opened 2021-04-14
An issue with on_auth_m5


viktar-instinctools:
### Environment

- the latest docker image from https://github.com/vernemq/docker-vernemq
- .net core application in the same docker
- webhooks are configured in vernemq.conf and vernemq can send requests to .net core application

### Expected behaviour

-  vernemq handles response of on_auth_m5 webhook

### Actual behaviour

- Hello all,
I am trying to understand if vernemq suits us, but I have faced with an issue which I could not solve.
According to the [documentation ](https://docs.vernemq.com/plugindevelopment/webhookplugins#on_auth_m5 ) for web_hooks the proper response for on_auth_m5 is

```
{
   "modifiers":{
      "properties":{
         "p_authentication_data":"QVVUSF9EQVRBMQ==",
         "p_authentication_method":"AUTH_METHOD"
      },
      "reason_code":0
   },
   "result":"ok"
}
```
But looks it's not a proper structure of response according to the logs of vernemq container

{p_authentication_data => <<"AUTH_DATA1">>,p_authentication_method => <<"AUTH_METHOD">>},reason_code => 0}} in vmq_mqtt5_fsm:check_enhanced_auth/2 line 789

13:51:25.837 [error] Ranch listener {{172,17,0,2},1883} terminated with reason: no case clause matching {ok,#{properties => #{p_authentication_data => <<"AUTH_DATA1">>,p_authentication_method => <<"AUTH_METHOD">>},reason_code => 0}} in vmq_mqtt5_fsm:check_enhanced_auth/2 line 789

Then I've changed the structure of response to 

```
{
  "properties":{
         "p_authentication_data":"QVVUSF9EQVRBMQ==",
         "p_authentication_method":"AUTH_METHOD",
         "reason_code":0
      },
   "ok":"ok"
}
```
After this I checked logs of vernemq container:

13:56:47.871 [error] CRASH REPORT Process <0.2290.0> with 0 neighbours crashed with reason: no case clause matching undefined in vmq_webhooks_plugin:handle_response/3 line 735

13:56:47.872 [error] Ranch listener {{172,17,0,2},1883} terminated with reason: no case clause matching undefined in vmq_webhooks_plugin:handle_response/3 line 735

What is the proper response for on_auth_m5?

ioolkos:
huh, good call @viktar-instinctools, thanks. Seems like the MQTT session and the hook do not agree on the format of the properties here.
Let me understand and possibly fix this.

[...]

PR #1839, opened 2021-06-28
Fixed mqtt5 enhanced auth flow webhook payload return handling

B-1P:
## Proposed Changes

## Types of Changes

- [X ] Bugfix (non-breaking change which fixes issue #XXXX)

## Checklist

- [X ] I have read the `CODE_OF_CONDUCT.md` document
- [ X] All tests pass locally with my changes

## Further Comments

The return for webhook during enhanced auth flow for mqtt5 is not handled correctly. The case/matcher always fails. This PR fixes that.

### DIFF
--- a/apps/vmq_server/src/vmq_mqtt5_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt5_fsm.erl
@@ -787,7 +787,7 @@ check_enhanced_auth(#mqtt5_connect{properties=#{?P_AUTHENTICATION_METHOD := Auth
                              ?P_REASON_STRING], M)
         end,
     case vmq_plugin:all_till_ok(on_auth_m5, [UserName, SubscriberId, Props]) of
-        {ok, #{reason_code := ?SUCCESS,
+        {ok, #{reason_code := ?M5_SUCCESS,
                properties :=
                    #{?P_AUTHENTICATION_METHOD := AuthMethod} = Res}} ->
             EnhancedAuth = #auth_data{method = AuthMethod},


ioolkos:
@B-1P thanks! have you seen actual issues around that? I think `?SUCCESS` is part of the interface between plugin and broker, whereas `?M5_SUCCESS` would be the MQTT response code. So we likely don't want to change that but review pending.

B-1P:
> @B-1P thanks! have you seen actual issues around that? I think `?SUCCESS` is part of the interface between plugin and broker, whereas `?M5_SUCCESS` would be the MQTT response code. So we likely don't want to change that but review pending.

Thanks for the clarification around the difference between `?SUCCESS` and `?M5_SUCCESS`. From my testing, a json response as described in the [docs](https://msazure.visualstudio.com/One/_git/Azure-IoT-ConnectedVehicle-Core/pullrequest/4723477) with `"reason_code": 0`, does not actually match with `?SUCCESS`, but it does match with `?M5_SUCCESS`. This fix might not be exactly correct but there is definitely an issue here.

This is the issue I see without the change:
https://github.com/vernemq/vernemq/issues/1780

ioolkos:
@B-1P ok, thanks for bringing #1780 back to attention. I cannot immediately look into this but you're probably right that there's an inconsistency there.

lanbai1993:
@B-1P @ioolkos is this issue resolved now? i am now using vernemq version 1.12.3  and i want reject pub/sub operations through web hooks, but this response json
{\"result\":\"ok\",\"properties\":{\"reason_code\":135}}
is not correct.
can you guys give me the correct json response?
btw, if i use {\"result\":\"error\"}, the broker server occurs error like:
Ranch listener {{172,17,0,2},1883} terminated with reason: no case clause matching <<"error">> in vmq_webhooks_plugin:handle_response/3 line 799

lanbai1993:
> @B-1P @ioolkos is this issue resolved now? i am now using vernemq version 1.12.3 and i want reject pub/sub operations through web hooks, but this response json {"result":"ok","properties":{"reason_code":135}} is not correct. can you guys give me the correct json response? btw, if i use {"result":"error"}, the broker server occurs error like: Ranch listener {{172,17,0,2},1883} terminated with reason: no case clause matching <<"error">> in vmq_webhooks_plugin:handle_response/3 line 799

the hooks are auth_on_publish_m5 and  auth_on_subscribe_m5, not auth_on_m5

Issue #1633, opened 2020-09-16
Puzzle: Sometimes Retained messages seem to be skipped


ioolkos:
A user has sometimes seen Retained messages "skipped". The cluster will deliver the message on the normal subscription path, but for that topic, the cluster will still show the second-last retained message.

Looking for hypothesis/code path for this "puzzle". As the Retained message is missing on all cluster nodes, this doesn't seem to be related to synchronization (but it's not excluded).

Look into: Retain Cache, Retain SRV, Overload, State Synchronisation (SWC)

ioolkos:
We need to check for message_queue length in the mailbox of the Retain SRV process & and if this is in fact a growing mailbox, explore options to introduce concurrency.

ioolkos:
Ok, the above (mailbox overload) is confirmed while disk latency seems okay.
Looks like a retain_srv_sup_sup -> retain_svr_sup -> multiple supervised retain_srv processes might give concurrency. (of course needing a way to distribute the incoming messages over those processes).

ioolkos:
Suppose we `phash2` on the topic name to map this to a certain retain_srv.
The question is what a common overload case is though: a) retains on a lot of different topics, or b) high speed updates on the same topic

elduds:
I have seen this retain issue on a topic that is rarely updated (2-3 times per day), though the broker cluster is busy processing 10-15kpps. Difficult to detect on topics that are updated frequently so it may be happening on others.

coder4803:
We have been noticing these "skips" with message bursts. Last example was that there were three messages (qos 1) published by the **same client** to the **same topic**. Messages did go through in the order they were published but when the topic was subscribed afterwards, we got the second/middle one.

I am guessing that the root cause could be that multiple threads can handle messages published by a single client (to a single topic) and retaining those messages may finish/happen in different (unexpected) order.

ioolkos:
@coder4803 
thanks, jep. On the cause: I think this is an overload of the Retain server at the incoming side. Basically, filling up the backlog of that process. This is then stored via the metadata system, which also distributes the retained messages to all the other cluster nodes. (This replication is eventually consistent, in any case)

elduds:
Regardless of the fix, something that would be useful is some prometheus metrics around what's happening with the retain server. We are already graphing #msgs and bytes, however this doesn't give any insights into an overload situation.

It would be great to see CRUD metrics as well as the size of the backlog for the retain server to aid troubleshooting/alerting.

cheers

[...]

PR #1866, opened 2021-08-11
Fix retain caching

MarcVanOevelen:
## Proposed Changes

This commit fixes retain_cache corruption.

When publishing messages with the retain flag set, successive updates to the same topic sometimes result in
a corrupted retain cache. This occurs sporadically in a cluster or on a single node. 
This problem has also been reported in this issue :
 Puzzle: Sometimes Retained messages seem to be skipped #1633

### Problem analysis:

  Incoming retain messages immediately update the RETAIN_CACHE managed by  vmq_retain_srv.
  Every retain_persist_interval (default 1s) vmq_retain_srv forwards the recent updates to the vmq_metadata
  plugin. (= vmq_swc) to persist and distribute the updates to all cluster nodes.
  All nodes subscribed to the metadata will receive async notification events for those updates so that they can update
  their local RETAIN_CACHE and hence maintain sync.
  The node that triggered the update (originating node) is itself also subscribed to receive update events from the other nodes.
  Therefore the generic metadata plugin implementation also sends the updates back to the originating node
  which will then re-apply the update to it's RETAIN_CACHE.
  When in the mean time another incoming message had updated the cache, that update will be "skipped".
  
  So the root cause is the feedback event that results in a second update to the cache. 
  
###  Fix implemented in vmq_swc_store.erl:
  
  Skip update events to the originator.
  To avoid potential impact on other metadata tables, skipping is only performed for RETAIN_DB events.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #1633)

## Checklist

## Further Comments

Note that plumtree metadata plugin might also suffer from this issue,
but this pull request only fixes swc since that is now the preferred plugin.

### DIFF
--- a/apps/vmq_swc/src/vmq_swc_store.erl
+++ b/apps/vmq_swc/src/vmq_swc_store.erl
@@ -85,6 +85,7 @@
 
 %-define(DBG_OP(Str, Format), io:format(Str, Format)).
 -define(DBG_OP(_Str, _Format), ok).
+-define(RETAIN_DB, {vmq, retain}).
 
 start_link(#swc_config{store=StoreName} = Config) ->
     gen_server:start_link({local, StoreName}, ?MODULE, [Config], []).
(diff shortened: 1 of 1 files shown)


ioolkos:
@MarcVanOevelen that's very cool, thanks a lot. 
We need to test this at high input rates, to see if it fixes mailbox backlog buildup. But this seems to address a major flaw in the current design.

ioolkos (approved):
No exported functions changed, so looks directly mergeable and good to me.

PR #1879, opened 2021-09-09
Compat OTP 24 with cuttlefish fixed to 4cf5e1c

manifest:
#1879 with with cuttlefish fixed to `4cf5e1c`.


ioolkos:
uhg, it's hard to find a set of dependencies that works with all 22-24 (`rebar3_cuttlefish/cuttlefish` basically)

manifest:
Maybe it's time to deprecate some OTP version(s) or cuttlefish or both? :-)

ioolkos:
@manifest I kinda like the old approach of supporting 3 OTP versions :) 

I don't care too much about 22, but 23 I need (for the official builds). Also: there's the issue of plugins needing to be the same version as Verne. Yours might already support 24, but it may well be that not all do.

PR #1880, opened 2021-09-10
Fix SWC event broadcast regression

ioolkos:


ioolkos:
Note to self: (from a use case testing the fix)
- this is not a full fix
- issue is not reproducable on a local dev cluster (Verne nodes on same machine)
- only reproducable when nodes run on different servers
- issue shows as delayed syncronisation (regarding subscriptions)

PR #1885, opened 2021-09-20
WIP: RocksDB, Leveled

ioolkos:


codeadict:
How is this feature going @ioolkos ? Anything i can help with?

ioolkos:
@codeadict thanks for asking! :)
I'm thankful for any input/help and it's a good opportunity to come back to this for me. What I have not yet tackled is to integrate the RocksDB config (as seen in the example advanced config file in this PR) vals into the `vernemq.conf` file. To be honest, I'm not sure all will be needed, but OTOH if we expose some, we could as well just go for all.

Other than that, we need to update everything in the branch here, make sure it runs on OTP 24, make sure the test suites go through with a RocksDB build.

There's also full Leveled support hidden in the PR :) Same as above applies, although I don't even think we need to configure much via `vernemq.conf` file.

codeadict:
Sounds good @ioolkos , doesn't seem like a hard task, although there is a lot of details in RocksDB config. Should I bring `vernemq:feature-rocksdb-messagestore` up to date with master? Then i will start working on my own based on this one

ioolkos:
> Should I bring `vernemq:feature-rocksdb-messagestore` up to date with master? Then i will start working on my own based on this one

Awesome :) yeah I think that's the best approach.

codeadict:
Alright, it is up to date here https://github.com/codeadict/vernemq/tree/feature-rocksdb-messagestore

ioolkos:
@codeadict I created the `new-feature-rocksdb-messagestore` branch here, including the commit from your branch.
For now, I'll keep RocksDB config in the advanced.config file; better to give users something to experiment.
Closing this PR in favour of new branch.

Issue #1915, opened 2021-12-06
Metrics for vmq webhook


iam-nagarajan:
### Environment

- VerneMQ Version: 1.12
- Runtime: Docker

We are missing metrics about how vmq webhook pools are used, the response latency from webhook endpoint. It will be helpful to have these metrics exposed so that we can fine tune the setup.

ioolkos:
Implemented by https://github.com/vernemq/vernemq/pull/1991

PR #1991, opened 2022-05-27
Collect metrics on webhook calls.

varnerac:
## Proposed Changes

Collect metrics per webhook type.

## Types of Changes

## Checklist

## Further Comments

I ran the webhooks suite and dialyzer. They pass. I haven't run the full VMQ suite because of clique issues.


varnerac:
This isn't ready to merge yet. However, I wanted to post what I have to see if this is the right approach. It addresses https://github.com/vernemq/vernemq/issues/1915

varnerac:
@ioolkos I think I should change this, and just add the webhook metrics here: https://github.com/vernemq/vernemq/blob/master/apps/vmq_server/src/vmq_metrics.erl#L681

It kind of conflates the `vmq_server` and `vmq_webhooks` model. I could also initialize all webhook metrics on VMQ web hooks startup. My initial strategy was to initialize the metrics on demand, because I wanted to a metric per webhook name (vs webhook "type"). The metrics system isn't really built for dynamically named metrics. So, I think the best route is to statically initialize the webhooks counters in the counters definition function in vmc_metrics. I'll have to leak the webhooks module in because of [met2idx](https://github.com/vernemq/vernemq/blob/master/apps/vmq_server/src/vmq_metrics.erl#L1161).

ioolkos:
@varnerac thanks, one thing unclear to me: did you try to use a `vmq_metrics_mfa` (as, for example, in the bridge plugin or SWC plugin) and found specific reasons against it?
The simplicity of the `vmq_metrics_mfa` approach let's you just export a `metrics/0` function from any custom metrics module, and it'd only be enabled when the plugin is active.
Have you found this too rigid for the Webhooks case?

varnerac:
@ioolkos I think the vmq_metrics_mfa is probably a better approach. I'll work on that.

varnerac:
@ioolkos This one use the metrics MFA callback

ioolkos:
@varnerac allright, thanks, looks good.

ioolkos (approved):

ioolkos:
This broke tests apparently:
```
rebar3 ct
[...]
===> Compiling apps/vmq_server/test/vmq_publish_SUITE.erl failed
apps/vmq_server/test/vmq_publish_SUITE.erl:8:10: can't find include file "../src/vmq_metrics.hrl"

apps/vmq_server/test/vmq_publish_SUITE.erl:742:19: record metric_def undefined
``

varnerac:
I'll patch

Issue #1982, opened 2022-04-27
QoS behavior bis


dometec:
### Environment

- VerneMQ Version: 1.12.3 (docker)
- OS: Fedora
- VerneMQ configuration (vernemq.conf) or the changes from the default: Set only ACCEPT_EULA / ALLOW_ANONYMOUS / UPGRADE_OUTGOING_QOS
- Cluster size/standalone: standalone

### Expected behaviour

I test locally the behaviour of VerneMQ using mosquitto clients when subscriber and publisher are using different QoS level.
What I'm expecting is resumed in "**3.8.4 Response**" in "**Non normative examples**" and "**Non normative comment**" of http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html: 

_Subscribing to a Topic Filter at QoS 2 is equivalent to saying "I would like to receive Messages matching this filter at the QoS with which they were published". This means a publisher is responsible for determining the maximum QoS a Message can be delivered at, but a subscriber is able to require that the Server downgrades the QoS to one more suitable for its usage._

So, I starter a new VerneMQ standalone server with:

`docker run --rm -it -p 1883:1883 -e "DOCKER_VERNEMQ_ACCEPT_EULA=yes" -e "DOCKER_VERNEMQ_ALLOW_ANONYMOUS=on" -e "DOCKER_VERNEMQ_UPGRADE_OUTGOING_QOS=off" --name vernemq vernemq/vernemq:1.12.3`

Subscribe with mosquitto_sub to a topic qith QoS 2:

> [dometec@localhost ~]$ mosquitto_sub -d -q 2 -t 'topic1'
> Client null sending CONNECT
> Client null received CONNACK (0)
> Client null sending SUBSCRIBE (Mid: 1, Topic: topic1, QoS: 2, Options: 0x00)
> Client null received SUBACK
> Subscribed (mid: 1): 2

And publish a message with QoS 1:

> [dometec@localhost ~]$ mosquitto_pub -d -t 'topic1' -m 'ciao' -q 1
> Client null sending CONNECT
> Client null received CONNACK (0)
> Client null sending PUBLISH (d0, q1, r0, m1, 'topic1', ... (4 bytes))
> Client null received PUBACK (Mid: 1, RC:0)
> Client null sending DISCONNECT
> [dometec@localhost ~]$

Following the specification, I would aspect on subscriber a message with QoS 1.

### Actual behaviour

I get a message with QoS 2:

[...]

PR #2001, opened 2022-06-12
Fix outgoing QoS upgrades

ioolkos:
Don't write QoS 0 messages to messages store. (might introduce option to enable that)

Fixes https://github.com/vernemq/vernemq/issues/1982


ioolkos:
This also fixes outgoing (resulting) QoS to full spec compliance:

| Pub/Sub | 0 | 1  | 2 |
| ------ |:------:| ---:| --:|
| 0      | 0      | 0   | 0 |
| 1      | 0      |   1 | 1 |
| 2      | 0      |   1 | 2 |

instead of always upgrading to the subscription QoS (which should only happen if `upgrade_outgoing_qos` is enabled):

| Pub/Sub | 0 | 1  | 2 |
| ------ |:------:| ---:| --:|
| 0      | 0      | 1   | 2 |
| 1      | 0      |   1 | 2 |
| 2      | 0      |   1 | 2 |

Issue #1670, opened 2020-11-05
mqtt_connect_timeout configuration is invalid


shenfeng07:
### Environment

- VerneMQ Version: 1.11.0
- OS: CentOS 8.0
- Erlang/OTP version (if building from source):
- VerneMQ configuration (vernemq.conf) or the changes from the default :
listener.ssl.default = 0.0.0.0:8883
listener.ssl.allowed_protocol_versions = 3,4,5
listener.ssl.require_certificate = on
mqtt_connect_timeout = 30000
- Cluster size/standalone:

### Expected behaviour

Set mqtt_connect_timeout to 30000 for allows more time for constrained devices to process certificates for establishing SSL connections

### Actual behaviour

I tried setting mqtt_connect_timeout to 30000，but broker still close connection after 5s.

![1](https://user-images.githubusercontent.com/29396363/98187359-b420f880-1f4b-11eb-8b28-39e231c24a15.png)
![2](https://user-images.githubusercontent.com/29396363/98187367-b71be900-1f4b-11eb-9028-0943131a2af7.png)

ioolkos:
Thanks @shenfeng07 
I think the `mqtt_connect_timeout` was only implemented half-way. (so it's possible this does not work with TLS). We should address this and have something like the `handshake_timeout` and `ssl_handshake_timeout` as RabbitMQ has.

rbino:
@ioolkos bumping this because we just ran into the same issue. On the client side we were using an ESP32 device (a microcontroller) which took just more than 5 seconds to perform the TLS handshake, which resulted in a timeout.

I'm willing to send a PR to try to fix this, but I need some help with the configuration side. I already determined that the `handshake_timeout` option should be passed [around here](https://github.com/vernemq/vernemq/blob/37283c78201cb444ab9bad8f492dc5131ae92d7b/apps/vmq_server/src/vmq_ranch_config.erl#L137) to Ranch.

Now, the `tcp` transport actually has a [no-op handshake](https://github.com/ninenines/ranch/blob/48476b475b4fed098ba75ab7574d3b451231dede/src/ranch_tcp.erl#L128) while the `ssl` transport [actually uses](https://github.com/ninenines/ranch/blob/48476b475b4fed098ba75ab7574d3b451231dede/src/ranch_ssl.erl#L180) the timeout (and it's the one that probably fail in the issue here).

So my doubts are:
- Which transports should support the `handshake_timeout` option?
[...]

PR #2021, opened 2022-08-10
Expose handshake_timeout option in listeners

rbino:
## Proposed Changes

Ranch had a default value of 5 seconds for the handshake timeout. This sometimes caused constrained devices to fail the connection since they couldn't complete the SSL handshake in that amount of time.

This commit allows configuring the handshake timeout to arbitrary values to support such use cases.

This also includes a small internal refactoring, renaming the old `transport_opts` to `socket_opts`, since that's what they actually are, and introducing new base `transport_opts` that depend on the ranch transport.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #1670)

## Checklist

### DIFF
--- a/apps/vmq_server/priv/vmq_server.schema
+++ b/apps/vmq_server/priv/vmq_server.schema
@@ -661,6 +661,69 @@
                                                                                   hidden
                                                                           ]}.
 
+%% @doc Set the TLS handshake timeout. This represents the maximum time that can
+%% be spent completing the TLS handshake for the protocol.
+%%
+%% It can be useful to set this to a higher value when clients are running on
+%% constrained devices (e.g. microcontrollers) which can take more than 5 seconds
+%% (the default value) to complete the TLS handshake.
+%%
+%% This can be specified either on the protocol level:
+%%
+%%     - listener.ssl.tls_handshake_timeout
+%%     - listener.wss.tls_handshake_timeout
+%%     - listener.vmqs.tls_handshake_timeout
+%%     - listener.https.tls_handshake_timeout
+%%
+%% or on the listener level:
+%%
+%%     - listener.ssl.my_ssl_listener.tls_handshake_timeout
+%%     - listener.wss.my_wss_listener.tls_handshake_timeout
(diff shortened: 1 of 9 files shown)


ioolkos:
@rbino Thanks.
- Can you take out the documentation for TCP listeners in the schema file (unless, of course, you see a use case.. but as you mentioned to go into no-op anyway)

Just FYI: We're currently fixing the tests, and some fail on Github actions anyway.

rbino:
@ioolkos should I just delete the documentation or should I delete everything (i.e. keep `handshake_timeout` only on SSL transports like `mqtts`, `wss`, `https`...)?

ioolkos:
@rbino apologies, missed your questions above. I'd keep the `handshake_timeout` option only for SSL transports, yes.

rbino:
@ioolkos updated. I also updated the merge description, in particular I added:

This also includes a small internal refactoring, renaming the old transport_opts to socket_opts, since that's what they actually are, and introducing new base transport_opts that depend on the ranch transport.

ioolkos:
@rbino do you want to keep this Draft?

rbino:
@ioolkos marked as Ready for Review. I tried running tests locally but I've found some unrelated failures - are these the same which are failing in the CI?

rbino:
@ioolkos btw I ran a VerneMQ image built from my branch in a full Astarte system against the hardware that was failing and I verified this actually works and fixes the problem when increasing the timeout to 30 seconds using environment variables

ioolkos (approved):

ratonlaveur38:
@rbino  , I am planning to give it a try with our device in China => usually fails 5s timeout.

After compilation, I was checking the vernemq,conf generated by `make rel` and could not find the new parameters. `listener.https.tls_handshake_timeout` etc ... 

Not sure how this file is generated and might be that I missed something (in this case discard the comment  :-) )

ioolkos:
@ratonlaveur38 those options are added as `hidden` configs so as to not clutter the conf file. (and most users will not set them)
You can just add them to the conf file yourself when you need to set them.

Issue #1412, opened 2020-01-15
RPC failed: timeout when tracing client


thewritingdude:
When trying to trace a client, after a time varying from 5 seconds to a minute, the trace will inevitably cause an error like

RPC to 'VerneMQ@10.0.0.7' failed: timeout

### Environment

- VerneMQ Version: 1.6.2
- OS: Amazon Linux 2
- VerneMQ configuration (vernemq.conf) or the changes from the default:
Added vmq_diversity postgres, changed nodename to VerneMQ@10.0.0.7 and listener.vmq_clustering to 10.0.0.7:44053
- Cluster size/standalone: 2

### Expected behavior

I should be able to trace a client by issuing the command:

vmq-admin trace client client-id=client-id1 --rate-interval=10 --rate-max=1000

### Actual behaviour

The trace will go on for some time and then exit with the error: RPC to 'VerneMQ@10.0.0.7' failed: timeout

ioolkos:
@thewritingdude are those nodes overloaded? (look at CPU). 
BTW, why do you use an old release like 1.6.2?

thewritingdude:
CPU usage are in the ballpark of 4-5% (for each node). 
Is there a known issue with old versions? Updating is not a problem, but if it can be avoided it would be better.

thewritingdude:
By the way, I updated to 1.10.1 and same result, after some time the same problem as described above.

Sleepingbug:
@scazzumvivendi Did you solve it ?

Sleepingbug:
@scazzumvivendi Did you solve it ? I have a same question. version: 1.11.0

scazzumvivendi:
Hi @Sleepingbug , sorry for the delay.

I found an issue on the file vmq-admin. 
I made a simple pull request (mentioned above), try that fix if it works also for you.

mths1:
Hi @scazzumvivendi 

I commented on your Pull Request. I am able to reproduce on both RHEL and Ubuntu (without EC2). The problem seems to be in what the change user function does. Maybe you can rework your pull request accordingly? This is not limited to EC2 and actually should hit anyone who is not working with the vernemq user (or whatever user was choosen)

scazzumvivendi:
@mths1 cool, thanks for the tip; it clearly looks better than my hack :)

The pull request is closed, you can open a new one if you want.

mths1:
I created a pull request that should fix the root cause. Feel free to test, comment etc.

PR #2048, opened 2022-10-18
Fix rpc timeout

scazzumvivendi:
## Proposed Changes

## Types of Changes

- [X] Bugfix (non-breaking change which fixes issue #1412)

## Checklist

- [X] Any dependent changes have been merged and published in related repositories

## Further Comments

considered, etc.

### DIFF
--- a/files/vmq-admin
+++ b/files/vmq-admin
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # -*- tab-width:4;indent-tabs-mode:nil -*-
 # ex: ts=4 sw=4 et
 


ioolkos:
@scazzumvivendi Thanks! can you explain, though: why does switching `sh` to `bash` fix a timeout? what OS are you on?

scazzumvivendi:
@ioolkos the switch is needed for the change in the other commit, i.e. from 
```
case $1 in
```
to
```
ARG=($1)
case $ARG in
```
so that in the EC2 machine $1 is interpreted as the first parameter, and not as all the parameters. I don't know how to pinpoint this, but it works and it doesn't introduce regressions (that I am aware of).

ioolkos:
@scazzumvivendi by the way: for what VerneMQ version (and self-compiled or subscription binary?) do you see this RPC timeout?
In general, I think it's best to keep the shell at `sh` which links to the system default, but I wondered what you actually use.

scazzumvivendi:
I'm using version 1.12.1 self-compiled, and installed on an Amazon Linux (EC2) machine.

It's OK if you leave this hanging and not merged on master, as it can be simply patched if anyone has this issue (which, I iterate, is caused by bash on my EC2 instance, and not by VerneMQ itself).

ioolkos:
@scazzumvivendi got, it, let's do it like that. I think we also change to bash in the package builds. Thanks for sharing your insights, in any case, I wasn't aware about this on EC2!

mths1:
I was also running into this problem. The root cause is the "check_user" function, as can be easiliy demonstrated. Just add a echo $1 before and after the check_user

echo "a $1"
# Make sure the user running this script is the owner and/or su to that user
check_user "$@"
echo "b $1"

and also a 

        echo "INFINITY"
        $NODETOOL rpc_infinity vmq_server_cli command $SCRIPT $@;;
    *)
        echo "NON INFINITY"

you'll see that when you run the script with the vernemq user everything works smooth $1 is "trace" as expected, and then you will see INFINITY. If the check_user functions does a su the script is called again, but due to some escaping $1 is now trace client client_id and this you'll see NON INFINITY. A quick fix would be to change 
case $1 in "trace"
to 
case $1 in trace*  

or look deeper into what check_user does. This is not related to EC2 or something, I can reproduce this on a local machine with both Ubuntu and RHEL.

mths1:
Changing

`exec su - $RUNNER_USER -s $RUNNER_SCRIPT_DIR/$RUNNER_SCRIPT -- "$ESCAPED_ARGS"`
to
`exec su - $RUNNER_USER -s $RUNNER_SCRIPT_DIR/$RUNNER_SCRIPT -- $ESCAPED_ARGS`

in env.sh should also fix the issue (and is actually the real root cause). The parantheses cause all arguments to be passed as one. No idea about what side effects that might cause, though, given that that code has been there for a while.

ioolkos:
@mths1 okay, thanks for analyzing this! I'll have to dig into this and check why we wanted to pass all arguments as one.

PR #2054, opened 2022-11-13
Add a default rule for all clients

omarmohamedkh:
## Proposed Changes

Each `client_id` needs to be known to add it to the ACL database, so to overcome this problem this patch adds a special client id `*` which will match all clients for a specific `username` and `password` only if there is no `client_id` for the current client in the database.

## Types of Changes

- [x] New feature
- [x] Cosmetics

## Checklist


ioolkos:
@omarmohamedkh Thanks, so this would basically add a catch-all Publish and Subscribe ACL than one can configure by adding a user with the Clientid "*" to the Database. The Client would still have to provide a Username and a Password as well as the ClientID (not "*" though!).

The benefit of this is that we'd have a default set of ACLs. The drawback would be the same: we only have 1 set of ACLs, no further "groups" of clients.

Folks have always adapted the Lua scripts in various ways. This was always intended.

We have to carefully consider what we want to be maintained and documented common functionality in the scripts. I'll try to think about it and make a decision quickly. In any case, you'd have shown a way to achieve a "catch-all" should any other users have the exact same need. (I'd be curious to know btw).

Anyone interested, please comment!

Ashtonian:
I think this is pretty useful. We have a legacy IoT fleet that shares credentials per device per tenant but obviously each device needs its own client id and we would like to avoid _having_ to preregister each device. Additionally we have a ingest service with multiple pods that dynamically generate their client id on the fly to be unique as a way to force a new session and pre-registering those also seems excessive.

emqx auth plugin simplifies this by adding a required columns configuration that defaults with 'client_id' but removing that column ignores it.

ashtonian:
Any additional feedback?

ioolkos:
@ashtonian Thanks for the reminder. I just see I made a promise on a quick decision... ugh, apologies.
Will re-check, it cannot break existing usages or lower security.
I'd like to note one thing again, though: the scripts as proposed here can already be used if needed. That is, you can replace the original scripts with the ones here.

ashtonian:
Any chance this can get merged?

PR #2067, opened 2022-12-28
Update Makefile for Darwin for OpenSSL Directory

bopbi:
## Proposed Changes

used brew prefix path to provide the correct openSSL dir

## Types of Changes

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/apps/vmq_passwd/c_src/Makefile
+++ b/apps/vmq_passwd/c_src/Makefile
@@ -3,7 +3,7 @@ CFLAGS_EXTRA =
 
 ifeq ($(shell uname -s), Darwin)
 	# OSX with homebrew
-	OPENSSL_DIR ?= /usr/local/opt/openssl
+	OPENSSL_DIR = $(brew --prefix openssl)
 	CFLAGS_EXTRA += -L${OPENSSL_DIR}/lib -I${OPENSSL_DIR}/include
 endif
 
--- a/apps/vmq_passwd/c_src/Makefile
+++ b/apps/vmq_passwd/c_src/Makefile
@@ -3,7 +3,15 @@ CFLAGS_EXTRA =
 
 ifeq ($(shell uname -s), Darwin)
 	# OSX with homebrew
-	OPENSSL_DIR ?= $(brew --prefix openssl)
+    HAS_BREW := $(shell command -v brew;)
+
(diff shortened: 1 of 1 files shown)


apps/vmq_passwd/c_src/Makefile
```diff
@@ -3,7 +3,7 @@ CFLAGS_EXTRA =
 
 ifeq ($(shell uname -s), Darwin)
 	# OSX with homebrew
-	OPENSSL_DIR ?= /usr/local/opt/openssl
+	OPENSSL_DIR = $(brew --prefix openssl)
```
codeadict, line 6: We should keep the ?= to support osx without homebrew

bopbi: @codeadict thank you, PR updated

apps/vmq_passwd/c_src/Makefile
```diff
@@ -3,7 +3,7 @@ CFLAGS_EXTRA =
 
 ifeq ($(shell uname -s), Darwin)
 	# OSX with homebrew
-	OPENSSL_DIR ?= /usr/local/opt/openssl
+	OPENSSL_DIR ?= $(brew --prefix openssl)
```
codeadict, line 6: I thought a little bit more and tested with self compiled open SSL instead of brew, would be good to keep the default install location in case the user uses self compiled homebrew or macports. Something like this should work:

```suggestion
    HAS_BREW := $(shell command -v brew;)

	ifdef HAS_BREW
		OPENSSL_DIR ?= $(shell brew --prefix openssl)
	endif

	# Default dir
	OPENSSL_DIR ?= /usr/local/opt/openssl

```

apps/vmq_passwd/c_src/Makefile
codeadict: @bopbi - See comment below, we should be able to merge after that with @ioolkos approval

apps/vmq_passwd/c_src/Makefile
bopbi: hmm nice idea, i am using your changes

codeadict (approved):
Lgtm, tested on my mac

ioolkos (approved):

PR #2086, opened 2023-02-14
Introduces API Key for all exposed HTTP endpoints

mths1:
## Proposed Changes
- Extend the scope of API keys to all http endpoints (metrics, health...)
- Generalize the auth functionality a bit, which will allow jwt authentication later on
- Introduce a expiry date that will support key rotation
- Auth for each endpoint can be turned on/off via verne.conf

This pull request is for an initial review and not yet fully finalized.

## Types of Changes

## Checklist

- [ ] I have read the [CODE_OF_CONDUCT.md](https://github.com/vernemq/vernemq/blob/master/CODE_OF_CONDUCT.md) document

## Further Comments

considered, etc.

### DIFF
--- a/apps/vmq_server/src/vmq_http_config.erl
+++ b/apps/vmq_server/src/vmq_http_config.erl
@@ -13,7 +13,7 @@
 %% limitations under the License.
 
 -module(vmq_http_config).
--export([config/1]).
+-export([config/1, auth_mode/2]).
 
 -callback routes() -> [{string(), atom(), any()}].
 config(Opts) ->
@@ -41,3 +41,29 @@ config([HttpModule | Rest], Routes) when is_atom(HttpModule) ->
     end;
 config([], Routes) ->
     Routes.
+
+auth_mode(Req, Module) ->
+    Reg = maps:get(ref, Req),
+    Scheme = binary_to_atom(maps:get(scheme, Req)),
+    Opts = ranch:get_protocol_options(Reg),
+    Global = maps:get(
+        Module,
+        vmq_config:get_env(vmq_server, http_modules_auth, #{}),
+        undefined
+    ),
+    ListenerName =
+        case lists:keyfind(listener_name, 1, maps:get(opts, Opts)) of
+            false -> undefined;
+            Tuple -> list_to_atom(element(2, Tuple))
+        end,
+    Local = maps:get(
+        {Scheme, ListenerName, Module},
+        vmq_config:get_env(vmq_server, http_listener_modules_auth, #{}),
+        undefined
+    ),
+
+    case {Local, Global} of
+        {undefined, undefined} -> undefined;
+        {undefined, G} -> G;
+        {L, _} -> L
+    end.
--- a/apps/vmq_server/src/vmq_status_http.erl
+++ b/apps/vmq_server/src/vmq_status_http.erl
@@ -16,7 +16,7 @@
 -behaviour(vmq_http_config).
 -include("vmq_metrics.hrl").
 
--export([routes/0]).
+-export([routes/0, is_authorized/2]).
 -export([node_status/0]).
 -export([
     init/2,
@@ -47,6 +47,14 @@ content_types_provided(Req, State) ->
         Req,
         State
     }.
+
+is_authorized(Req, State) ->
+    AuthMode = vmq_http_config:auth_mode(Req, vmq_status_http),
+    case AuthMode of
+        "apikey" -> vmq_auth_apikey:is_authorized(Req, State, "status");
+        "noauth" -> {true, Req, State};
+        _ -> {error, invalid_authentication_scheme}
+    end.
 
 terminate(_Reason, _Req, _State) ->
     ok.
--- a/apps/vmq_server/priv/vmq_server.schema
+++ b/apps/vmq_server/priv/vmq_server.schema
@@ -233,7 +233,7 @@
 %% transmitted simultaneously. This includes messages currently going through handshakes
 %% and messages that are being retried. Defaults to 20. Set to 0 for no maximum. If set
 %% to 1, this will guarantee in-order delivery of messages.
-%% Note: for MQTT v5, use receive_max_client/receive_max_broker to implement 
+%% Note: for MQTT v5, use receive_max_client/receive_max_broker to implement
 %% similar behaviour.
 {mapping, "max_inflight_messages", "vmq_server.max_inflight_messages", [
                                                                         {default, 20},
@@ -348,6 +348,7 @@
                                                       {datatype, string},
                                                       hidden
                                                      ]}.
+
 {translation, "vmq_server.http_modules",
  fun(Conf) ->
          S = cuttlefish:conf_get("http_modules", Conf),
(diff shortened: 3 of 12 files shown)


ioolkos:
@mths1 thanks, I'm not against this if you feel it adds value and maybe some added security.
Note that:

- the HTTP endpoints were never meant to be exposed to 'world'
- the Status page was never meant to grow into a management GUI

Since the HTTP API is obviously "management", it was the only endpoint using the api key.

apps/vmq_server/src/vmq_status_http.erl
```diff
@@ -48,6 +48,19 @@ content_types_provided(Req, State) ->
...
 
+is_authorized(Req, State) ->
+    case
+        maps:get(
+            auth_status_http,
+            vmq_config:get_env(vmq_server, http_modules_auth, undefined),
+            undefined
+        )
+    of
+        "apiley" -> vmq_auth_apikey:is_authorized(Req, State, "status");
```
codeadict, line 59:
```suggestion
        "apikey" -> vmq_auth_apikey:is_authorized(Req, State, "status");
```

codeadict:
I kinda like the idea, i've been myself thinking about a management API that does not use clique behind the scenes that we could use for a real full featured management UI. Its all in ideas but i do not have a full design written

ioolkos:
Yes, most users would keep the API key off for Prometheus and health enpoint (which is checked by loadbalancers).

I totally agree with your idea of an extended HTTP API, @codeadict. The HTTP wrapper works and keeps the CLI and HTTP admin interface nicely in sync. But it's sometimes a little difficult to explain to users.

There was a small experiment here (as an example on how to expose more of internal functions):
https://github.com/boboxiaodd/web_agent

And we might check https://github.com/gojek/vernemq and see whether they have done something similar (along with other fixes they might have). It seems we have to actively check and chase those contributions as they do not open PRs to  VerneMQ upstream, at least so far.

mths1:
Hi @ioolkos, @codeadict 

I agree that you can setup an environment where you do not need to protect your scrape endpoints or your metrics endpoints (or not even can do, for some LBs). The pull request has two aspects:

- Maybe I have gone through too many security audits recently, but there surely is a trend towards requiring "secure all endpoints, so that they could be exposed to the world, even if they are not". In real life, misconfigured network setups do happen, too broad firewall rules as well etc. Prometheus can handle secured scrape endpoints, so it is up to the user if he wants to enable it or not (and not everyone is running k8s).

- Key rotation is a requirement in many security handbooks as well (if you are using API keys). The proposed change will allow to enforce key rotation, or enable to give external personell/contractors, or specific applications access to the mgmt api for a defined period of time.

ioolkos:
@mths1 we'll certainly merge this to tick off your items from the list.
This does not stand in the way of a more extended HTTP in any case.

Let's think about what tests to add to ensure we don't break features later. (test for key rotation, as a minimum).

mths1:
One question: Would a nicer way of mapping HTTP Modules to HTTP(S) endpoints desirable? 

I just figured out how to set one http module to some other endpoint (e.g. having mgmt api on 444, and status on 443) which was non-obvious and also just works for one endpoint. It also includes removing it from http_modules so that the default loader is not loading it anway. Wouldn't it be nicer to have http_modules as an option on listener.http(s).[$name] instead of as an global config param?

ioolkos:
> Wouldn't it be nicer to have http_modules as an option on listener.http(s).[$name] instead of as an global config param?

I guess we can do this, yes. We should have some sort of backward compatibility, though. If the default config is there, everything should just work nicely on that default port (`8888`).

ioolkos:
looks functionally complete as per the original goal. Hit the ready button when you're ready, @mths1  :)

mths1:
Working on it. Cuttlefish is driving me crazy... As we now can set the module per listener I also wanted to have the auth mode per listener. I tried to be smart and implement something in cuttlefish in the lines of listener.http.$name.http_modules.$module.auth.mode but whatever I try falls apart soon after I thought I made some progress...

ioolkos:
I've never seen that double naming structure. 
Documentation goes nowhere near this:
https://github.com/basho/cuttlefish/wiki/Cuttlefish-for-Erlang-Developers
As for being Crazy-by-Cuttlefish... I know this feeling too well :)

apps/vmq_server/src/vmq_http_config.erl
```diff
@@ -41,3 +41,29 @@ config([HttpModule | Rest], Routes) when is_atom(HttpModule) ->
...
+auth_mode(Req, Module) ->
+    Reg = maps:get(ref, Req),
+    Scheme = binary_to_atom(maps:get(scheme, Req)),
+    Opts = ranch:get_protocol_options(Reg),
+    Global = maps:get(
+        Module,
+        vmq_config:get_env(vmq_server, http_modules_auth, #{}),
+        undefined
+    ),
+    lager:info("fff ~p", [maps:get(opts, Opts)]),
```
ioolkos, line 54: There's a couple of forgotten development info logs, let's take them out.

mths1: Taken out

ioolkos:
@mths1 I just noticed that `vmq_schema_SUITE` does not pass locally, I suppose neither does it for you?

`➜  vernemq git:(2086) rebar3 ct --suite apps/vmq_server/test/vmq_schema_SUITE.erl`

mths1:
Tests case should run through now. vmq_schema_SUITE never worked for me (see different pull request) so I didn't notice that it breaks more than before...

ioolkos:
> Tests case should run through now. vmq_schema_SUITE never worked for me (see different pull request) so I didn't notice that it breaks more than before...

I see... thanks. Sometimes we got to fix the tests not the code :)

ioolkos (approved):

mths1:
Added the missing config for https and improved cli output. Just tested it once more. Open for further testing. The documentation is here:
https://github.com/vernemq/vmq-docs/pull/57

This pull request is a prerequisite for the http_pub module and an upcoming jwt plugin.

ioolkos:
@mths1 :+1: thanks, will test.

ioolkos:
The following test case breaks, due to the changes in `vmq_http_config` (I think)

```
➜  vernemq git:(2086) ✗ rebar3 ct --suite apps/vmq_server/test/vmq_metrics_SUITE.erl --case simple_prometheus_test
```

ioolkos:
@mths1 thanks. I wonder about a thing, though: if the default is "no_auth" for HTTP listeners (which is good for backwards compat), we probably don't want to trip over a dynamically configured listener that does not give the option.

mths1:
Currently, live reconfiguration is not possible, yes. But I am also not sure if HTTP listeners can be started with vmq-admin. A safer default would be to have the most restrictive and make it less restrictive on a per listener level. This would break backwards compatability.

ioolkos:
> But I am also not sure if HTTP listeners can be started with vmq-admin

Good point, it's not meant to be exposed. (even if it's easily possible internally, as the `vmq_server_cmd` shows).
So we can leave the defaults as is. An HTTP listener started via config file will be of type `no_auth` in case Verne does not find a setting for it.
Thanks again for a great contribution, @mths1! :)

ioolkos (approved):

PR #2088, opened 2023-02-18
Allow more fine-grained control of which modules are loaded per listener

mths1:
## Proposed Changes

This change will allow to set http_modules as a per-listener setting. This allows to seperate, for example, the management API from the status and health pages and enables e.g. more fine-granular firewall settings. In case there are no per-listener settings, the default values are used.

For example, the default listener could be limited to the health and status modules by setting listeners.https.default.http_modules = vmq_status_http, vmq_health_http

## Types of Changes

## Checklist

- [X] I have read the [CODE_OF_CONDUCT.md](https://github.com/vernemq/vernemq/blob/master/CODE_OF_CONDUCT.md) document

## Further Comments

considered, etc.

### DIFF
--- a/apps/vmq_server/src/vmq_ranch_config.erl
+++ b/apps/vmq_server/src/vmq_ranch_config.erl
@@ -318,18 +318,24 @@ protocol_opts(cowboy_clear, Type, Opts) when
         stream_handlers => [vmq_cowboy_websocket_h, cowboy_stream_h]
     });
 protocol_opts(cowboy_clear, _, Opts) ->
-    Routes =
-        case {lists:keyfind(config_mod, 1, Opts), lists:keyfind(config_fun, 1, Opts)} of
-            {{_, ConfigMod}, {_, ConfigFun}} ->
-                try
-                    apply(ConfigMod, ConfigFun, [])
-                catch
-                    E:R ->
-                        lager:error("can't setup HTTP modules due to ~p:~p", [E, R]),
-                        []
-                end;
-            _ ->
+    Apply = fun(Mod, Fun, Param) ->
+        try
+            apply(Mod, Fun, Param)
+        catch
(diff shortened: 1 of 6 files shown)


ioolkos:
@mths1 nice :)
There's currently still at least one Dialyzer fail in the "Check code style and quality" test.
You can always run Dialyzer locally by `rebar3 dialyzer` btw.

apps/vmq_server/src/vmq_ranch_config.erl
```diff
@@ -334,18 +334,24 @@ protocol_opts(cowboy_clear, Type, Opts) when
...
+    Apply = fun(Mod, Fun, Param) ->
+        try
+            apply(Mod, Fun, Param)
+        catch
+            E:R ->
+                lager:error("can't setup HTTP modules due to ~p:~p", [E, R]),
                 []
+        end
+    end,
+    {_, ConfigMod} = lists:keyfind(config_mod, 1, Opts),
```
ioolkos, line 346: This implies that we always have a `config_mod` and `config_fun` in Opts, otherwise we crash here. I guess that's the case via defaults, though.

mths1: Following the least-surprise-principle: Not configuring any (http/https) listener...

ioolkos: You're right, I don't see any need to compile empty routes, which is technically possible. I don't know why we originally allowed it.

```
(VerneMQ@127.0.0.1)9> cowboy_router:compile([{'_', []}]).
[{'_',[],[]}]
```
Let's keep your version.

ioolkos:
@mths1 let's add a changelog & merge this.

ioolkos (approved):

PR #2093, opened 2023-02-23
Feature/vmq http client

mths1:
## Proposed Changes

Implementation of a HTTP POST Interface so that HTTP(S) clients can POST messages to VerneMQ directly. The plugin provides the following features:

* Three different "formats" for publishing data:
    - Payload in body and MQTT Information in header
    - Payload in body and MQTT Information as query parameters
    - JSON Format

* Integrates nicely into VerneMQ authentication schemes
    - Can protect endpoint with API Key
    - Can post "on-behaf-of", in that case the integrated authorization/authentication plugins are used 
    - Supports mountpoints
    - Supports setting topic, QoS and retain flag
    - Supports UserProperties (for MQTT5 subscribers)

## Types of Changes

## Checklist

- [ ] I have read the [CODE_OF_CONDUCT.md](https://github.com/vernemq/vernemq/blob/master/CODE_OF_CONDUCT.md) document

## Further Comments

considered, etc.


mths1:
Not yet ready for review. A small preview of a plugin I am preparing. Let me know if you think this fits into the main VerneMQ distribution. 

It is a small interface for publishing data to VerneMQ via HTTP(S). Compared to what I found elsewhere, the main goal is to integrate into VerneMQ (authentication, mountpoints, metrics...). It is meant to be used as a bridge for occasionally sending data to VerneMQ via an integrated mechanism. It is not designed for a lot of clients, nor a lot of traffic.

ioolkos:
This sounds good to me, in principle. 

So, on-behalf-of would do authorization, but other otherwise not, ie a component could just post to all topics?
Or would we need a non-MQTT user concept? ("admin" user). So far we have avoided this.

mths1:
My current thoughts are:
"on-behalf-of":
- The pw/client_id/user are part of the request (in either body, header or query string). 
- The information is send through the authorization plugins. 
- If ok, it will publish on behalf of what-ever-was-in-there
- I guess typically you would at least use a own client_id for rest based clients (I even thought about enforcing that)

endpoint-authorization:
Not yet sure, what is better. I considered: 
- "God mode", maybe restricted to some topic tree (most likely not such a good idea security wise)
- "Fixed user": The endpoint uses a fixed (configured) pre defined user/client pair. We use the ACLs of that pair.
- Mapping API Key to user/client, so the API Key selects also the MQTT user/client

The last option is most likely the most flexible one. Once we have JWT there would also be the option to encode ACLs in the JWT token.

ioolkos:
@mths1 an "on-behalf-of" MVP looks like a good option. Then we can always add to it later.

mths1:
The module is now feature-complete. It is waiting for the api key pull request for finalization. Needs some tests cases, but manual tests go throuh fine. Any comments are welcome.

ioolkos:
@mths1 would you have a minimal example... let's say a minimal config of the plugin in the vernemq.conf and some sort of `curl -x POST` example for a message publish?

mths1:
I will push a new version later today or tomorow which is ready to test. I am still puzzled how to properly write test cases. Currently, I have some smoke tests (as in I can publish and it returns 200 ok), but also only with authorization disabled and without checking if the data really made it to verne.

Here is the documentation (80% finished), so you can see what you can test tomorrow :-)

https://github.com/mths1/vmq-docs/blob/docu/http_pub_plugin/configuration/http-pub.md

mths1:
Should be ready for testing, some notes:
* It does a auth_on_register for every push/pop => this works good enough, as long as you do not use a slow hash (as in slow by design, as bycrpt). In a later version of the plugin, we might want to cache sucessfull on_auth_register for a while. 
* I still lack a good way of unit test the authentication overrides (topic, qos...). I did this manually, but it is easy to miss something while refactoring code at the same time. 
* Is there an easy way of checking what has actually been posted?

ioolkos:
@mths1 awesome, let me get through the docs you provided, set things up locally, and then I can look into the PR and your above questions.

ioolkos:
I cannot yet make this work. If I create an API key with `vmq-admin api-key create` shouldn't I be able to use this in an authorization header? (I get a malformed header exception from Cowboy)

```
curl --request POST \
  --url https://127.0.0.1:3001/restmqtt/api/v1/publish \
  --header 'Authorization: Basic LRq6kKehiQqD6UKyImlG1AwbzOxHdiX' \
  --header 'Content-Type: application/json' \
  --data '{
	"topic": "T1",
	"user": "myuser",
	"password": "test123",
	"client_id": "myclient",
	"qos": 1,
	"retain": false,
	"payload": "asddsadsadas22dasasdsad",
	"user_properties": [{"a":"b"}]
}'
```

mths1:
Oh, this happend to me as well, frequently. Basic auth needs to have a ":" at the end and then base64 encode that stuff. The following should work 
 
echo -n 'api-key-here': | base64

It bothered me that much that I added the x-api-key header in the api-key-pull-request  just because of this. 

--header 'x-api-key: api-key-just-as-is' 

will work without the base64 encoding.

My personal recommendation: https://insomnia.rest/

ioolkos:
@mths1 Thanks, gets me a step further :)
I missed the point that I need to create an API key with a scope: `vmq-admin api-key create scope=httppub`

This gets me past the initial error (with your hints about base64) but then I have below exception:

```
2023-03-10 10:33:54.571 [error] <0.893.0>@vmq_http_pub:auth_conf:180 Cowboy stream 1 with ranch listener {{127,0,0,1},3001} and connection process <0.892.0> had its request process exit with reason: no case clause matching [] in vmq_http_pub:auth_conf/1 line 180
```
Maybe I missed some config setting.

I tried with the following minimal plugin config:

```
plugins.vmq_http_pub = on
vmq_http_pub.mqtt_auth.mode = on-behalf-of  
listener.http.http_pub = 127.0.0.1:3001
listener.http.http_pub.http_module.vmq_http_pub.auth.mode = apikey
listener.http.http_pub.http_modules = vmq_http_pub
```

mths1:
Could you try adding the auth-mode to the listener?
listener.http.http_pub.http_modules.vmq_http_pub.mqtt_auth.mode = on-behalf-of

?

ioolkos:
> listener.http.http_pub.http_modules.vmq_http_pub.mqtt_auth.mode = on-behalf-of

This works. 

Another discovered point: We should discuss whether we want to respect `allow_anonymous=on` setting for the given (MQTT) username/pw/client_id.

`allow_anonymous` lets you disable authentication completely. For consistency reasons, we might allow it here too.

mths1:
I added a new push. Meanwhile, I was able to add some actual tests that also check authorization/authentication. I'd say the test coverage is quite good now. I will play a little bit more with it and then move it into review state

ioolkos:
@mths1 will do more testing soon. Thanks for the added tests (and your patience :)

mths1:
Moved from draft to for review. Most likely there is, both feature-wise and coding style, plenty of room for improvement. At least from a feature point of view, it should be "good enough" and in case there is interest from the community it can always be improved.

apps/vmq_server/priv/vmq_server.schema
```diff
@@ -390,7 +390,7 @@
   fun(Conf) ->
      HTTPModules = cuttlefish_variable:filter_by_prefix("http_module",Conf),
  	 HTTPModulesAsAtom = lists:map(fun(Element) -> list_to_atom(lists:nth(2,element(1, Element))) end, HTTPModules),
-     KnownHttpModules = [vmq_metrics_http,vmq_http_mgmt_api, vmq_status_http, vmq_health_http],%
+     KnownHttpModules = [vmq_metrics_http,vmq_http_mgmt_api, vmq_status_http, vmq_health_http, vmq_http_pup],
```
ioolkos, line 393: vmq_http_pub

ioolkos:
@mths1 sure, improvements can always be incremental. As is, I think we still want to do a little initial refactoring: the supervisor is unused as is but we could use it handle metrics (or not use at all). As a minimum we should also add the PublishFun to the state of `vmq_http_pub` and not create one for every publish we do.  Stuff like that.

mths1:
@ioolkos: Did some changes, made the code look nicer. Let me know if you have any further comments at the moment. I am not actively working on this PR at the moment.

apps/vmq_http_pub/src/vmq_http_pub_app.erl
```diff
@@ -0,0 +1,27 @@
...
+
+%% Application callbacks
+-export([start/2, stop/1]).
+
+%%====================================================================
+%% API
+%%====================================================================
+
+start(_StartType, _StartArgs) ->
+    {ok, _} = application:ensure_all_started(vmq_passwd),
```
ioolkos, line 18: We should not assume any specific authentication plugin.

mths1: I took the boilerplate from the passwd and acl plugin. This seems to be a leftover...

ioolkos (approved):

PR #2113, opened 2023-03-11
Fix dev_n builds and listener overlays.

ioolkos:
This tries to fix an issue, where listeners would be added to the generate app. files while not visible in `vernemq.conf`.
This also fixes the `make dev0 dev1` quick generation of local dev nodes. (creating correct nodenames, ports and listeners).

In addition, I enable v5 protocol as a default for WS and TLS listeners too (MQTT TCP had this for a while).

@codeadict I have the impression that {include_default, "default"} directive in schema files somehow creates issues, and we end up with 2 identical listeners in the generated app. files.


codeadict (approved):
Lpoks great to me, yeah @ioolkos i noticed the double listener but couldnt figure the cause

PR #2123, opened 2023-03-22
Bugfix: Client-id is set to late on enhanced authentication flow

### LINKED ISSUE #1780: An issue with on_auth_m5

### LINKED ISSUE #2026: incorrect username and client_id payload given to on_auth_m5 webhook
### Environment

- VerneMQ Version: 1.12.5
- OS: Ubuntu 22.04
- Erlang/OTP version (if building from source): 24

### Expected behaviour
Webhook [documentation](https://docs.vernemq.com/plugin-development/webhookplugins#on_auth_m5) mentioned sample expected payload : 
```
{
    "username": "username",
    "mountpoint": "",
    "client_id": "client-id",
    "properties": {
      "p_authentication_data": "QVVUSF9EQVRBMA==",
      "p_authentication_method": "AUTH_METHOD"
    }
}
```

### Actual behaviour
username and client_id value are incorrect when arriving on endpoint : 
```
{
    "username": null,
    "mountpoint": "",
    "client_id": "undefined",
    "properties": {
        "p_authentication_data": "<auth_data>",
        "p_authentication_method": "EXTERNAL",
        "p_session_expiry_interval": 3600
    }
}
```

Is the above behaviour expected ?

mths1:
Bugfix: Client-id is set to late on enhanced authentication flow on_auth_m5). It expects an atom (e.g. success) and not a number.

## Proposed Changes

Fixes #1780 : The documentation states that reason_code shall be a number (e.g.), but it actually expects an atom (e.g. success instead of 0)

Fixes #2026: This is more tricky. The client_id is only retrieved and checked (or calculated) after the on_auth_m5, within check_connect / check_client_id. This breaks check_connect/check_client_id apart and moves the client_id check before the on_auth_m5. That way the client_id will be available to the hook. This, of course, changes the behaviour a little bit. 

## Types of Changes

- [X] Bugfix (non-breaking change which fixes issue #XXXX)

## Checklist

## Further Comments

considered, etc.


ioolkos:
@mths1 ouh, nice! I almost forgot about that one.

> This, of course, changes the behaviour a little bit.

In what way? I don't immediately see a behaviour change in the FSM, except the reshuffling and the 'invalid' user thing?

mths1:
This was triggered because reason_code is now "success" and not 0. I am not a big fan of magic numbers, so I thought we want to continue with success instead of 0 (which is a breaking change). Nontheless, I have the feeling that this code path is largely unused.

Currently, for backwards-compatability I still call on_auth_m5 even if the client_id is invalid. I'd propose to not call on_auth_m5 and just fail in such cases (which would also be a breaking change)

Furthermore, I suspect that there are more problems waiting to be discovered in that on_auth_m5 flow.

ioolkos:
@mths1 yes, `on_auth_m5` is largely unused for the simple reason that it enables users to implement their own enhanced auth flows. Which is something users don't often need or don't want to do. It is still an important feature, though, so I'm glad to battletest this more! And even if it's only for potential internal plugins.
We can certainly go for explicit response codes ('success'). Not yet sure about invalid client_ids, I guess it depends what the enduser wants to do in the hook implementation (let's say log that event).
No worries about breaking changes here, in any case.

mths1:
It seems to be such an obscure feature that I did not find an ready-to-use example somewhere on the web to test against... mosquitto-pub can send authentication method and data, but cannot simulate the whole flow...

PR #2132, opened 2023-04-01
WIP: Initial v2 HTTP API

ioolkos:
Beginning of a v2 HTTP API, without the full management capabilities of v1 but with better performance for specific scriptable tasks. (avoiding `VMQL`).

### To use:

1. Create a specific API key for the v2 endpoint:
`vmq-admin api-key create scope=api2`
You can also use the general management API key, without the `api2` scope. When you create an API key on a cluster node, it will be replicated to other cluster nodes and can be used for calls on all cluster nodes.

2. Add a specific API2 listener to `vernemq.conf`, and restrict it to handle API2 only:
```
listener.http.api2 = 127.0.0.1:8885
listener.http.api2.http_modules = vmq_http_v2_api
http_module.vmq_http_v2_api.auth.mode = apikey
```

`api2` is the name of the listener (you can choose another one). `vmq_http_v2_api` is the module of code we want that endpoint to use. Auth mode `apikey` tells that module to require an API key. (to allow calls without key, set to `noauth`)

3. Use the API key for calls against the v2 API:
```
curl 'http://DkvQif8cWiWax43iJS8Vq6NyUk6lOMjs@127.0.0.1:8885/api/v2/client?client_id=tester&topics'
{"status":"online","topics":{"bla":0,"bla/tu/#":0,"test":0}}%
```

Check whether a Client is known, online or offline:
`curl 'http://DkvQif8cWiWax43iJS8Vq6NyUk6lOMjs@127.0.0.1:8885/api/v2/client?client_id=tester'`

Check the same with a specific given mountpoint (you don't have to give the default mountpoint):
`curl 'http://DkvQif8cWiWax43iJS8Vq6NyUk6lOMjs@127.0.0.1:8885/api/v2/client?client_id=tester&mountpoint=other'`

Check as above, but also get the list of topics the client is subscribed to:
`curl 'http://DkvQif8cWiWax43iJS8Vq6NyUk6lOMjs@127.0.0.1:8885/api/v2/client?client_id=tester&topics'`

Disconnect a client:
`curl 'http://DkvQif8cWiWax43iJS8Vq6NyUk6lOMjs@127.0.0.1:8885/api/v2/disconnect?client_id=tester'`

Disconnect and cleanup (=delete) state:
`curl 'http://DkvQif8cWiWax43iJS8Vq6NyUk6lOMjs@127.0.0.1:8885/api/v2/disconnect?client_id=tester&cleanup'`.

### DIFF
--- a/apps/vmq_server/priv/vmq_server.schema
+++ b/apps/vmq_server/priv/vmq_server.schema
@@ -351,7 +351,8 @@
 
 %% @doc specifies the modules to be enabled by the http(s) listeners. Overriding this setting is possible for specific listeners.
 {mapping, "http_modules", "vmq_server.http_modules", [
-                                                      {default, "[vmq_metrics_http,vmq_http_mgmt_api, vmq_status_http, vmq_health_http]"},
+                                                      {default, "[vmq_metrics_http,vmq_http_mgmt_api, vmq_status_http, vmq_health_http,
+                                                                vmq_http_v2_api]"},
                                                       {datatype, string},
                                                       hidden
                                                      ]}.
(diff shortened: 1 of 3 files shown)


ioolkos:
There's user interest to test this (smallish) features, so I'm merging it for now. Incremental changes/improvements can always be added, based on feedback.

Issue #2097, opened 2023-02-24
[Proposal]: Simplify/improve the terminology around listener management


codeadict:
### Describe your issue or feature request

Currently, in VerneMQ CLI, the management of the listeners seems somewhat confusing. You can "stop" a listener but it doesn't really stop it since you cannot start the same Address/Port combination again because it will return`{already_started, Pid}` even if the `vmq_admin listeners show` command presents the listener as "stopped". In that case, one needs to restart the listener to really start it again.,

### Describe the solution you'd like

As a simplified, less confusing terminology, I propose that `vmq_admin listeners stop ...` really stops the listener, killing all the sessions, and removing the listener from the show output, and that way start can work again. Also, remove the `restart` command and replace it with:

* `vmq_admin listeners suspend ...`: Does not kill connections but suspends any new connections.
* `vmq_admin listeners resume ...`: Starts receiving connections again on a suspended listener, do nothing on others.

Probably remove the `vmq_admin listeners delete ...` command too since stop will signify this. 

### Code of Conduct

- [X] I agree to follow the VerneMQ's Code of Conduct

ioolkos:
The terminology was chosen a long time ago... I agree it lacks clarity.
`suspend/resume` is great. `start/stop` will work, maybe `start/terminate` too to avoid redefining `stop`.

After that it's more of a documentation issue. I don't think the commands are widely used in a scripted fashion (breaking things).  In any case, better command terminology helps new users. So, adding my +1. Thanks for the idea, @codeadict !

---
:point_right: Thank you for supporting VerneMQ: https://github.com/sponsors/vernemq
:point_right: Using the binary VerneMQ packages commercially (.deb/.rpm/Docker) requires a paid subscription.

codeadict:
Indeed, `start/terminate` is even cleaner.

codeadict:
I have a branch for this , will open PR in the next few day

PR #2133, opened 2023-04-04
Fix suspension/restart of listeners, use more of Ranch API directly.

ioolkos:
Adapt suspension/restart of listeners. Fix tests a little bit.
`vmq-admin listener stop address=127.0.0.1 port=1885 -k` is now just a deletion of a listener and kept for backwards compatibility.

This also prepares for a naming more in line with Ranch (PR hasn't done the actual renaming yet)
https://github.com/vernemq/vernemq/issues/2097


ioolkos:
@codeadict no renaming regarding the listener management here. You mentioned you had a branch on this, let me know when you have better ideas. I just needed fix for session termination in cluster leave.

codeadict:
@ioolkos - Sounds good not to have the rename for now, I can adapt my branch later and we can do the renaming there. I'll do some testing with this later, this issue had to do with lots of the test failures we have so this is a great change

ioolkos:
@codeadict merging this for now but let me know if you had other fixes. Thanks for your help so far!

Issue #1663, opened 2020-10-29
Broker drops messages for clients with offline queue > max_online_messages


elduds:
This is a continuation of the discussion on vernemq-users mailing list:

(https://groups.google.com/g/vernemq-users/c/H59b4ygcbm8/m/2z_9Ygv2AgAJ)

### Environment

- VerneMQ Version: 1.11.0
- OS: Ubuntu 18.04 amd64
- Erlang/OTP version (if building from source): all
- VerneMQ configuration (vernemq.conf) or the changes from the default
max_online_messages = 75000
max_offline_messages = 76276800

### Expected behaviour

When a client with a large number of messages in their offline queue (> max_online_messages) reconnects after some time, all offline messages should be reliably delivered to the client.

### Actual behaviour

When the client comes online, *all* of their offline messages are loaded into the online queue. If the number of messages in their offline queue exceeds the broker's max_online_message value, then they receive max_online_messages quantity of messages however the remaining messages are all dropped.

This effectively means that the size of the offline queue is constrained to the size of the broker's RAM (max_online_messages) rather than the size of the broker's disk (max_offline_messages).

ioolkos:
Thanks, @elduds.

We have a blog post here on improving startup time after a broker reboot: https://vernemq.com/blog/2019/07/31/vernemq-startup-time-improvements.html
This is not the same situation as described in the current issue, but somewhat related. It seems that there is one more optimisation we should do. We do batch or "stream" the delivery from the loaded online queue to the re-connecting consumer as described in the blog post, but we do not "stream" the loading of the online queue from the offline queue.

rtuin:
Hi, I'm running into exactly the problem described by @elduds.

How can this be solved within verneMQ, would the "stream" from offline into online queues be hard to implement?
And is there a sensible work-around?

ioolkos:
@rtuin thanks, the only existing workaround currently is that `max_online_messages` is not configured to a lower value than `max_offline`.
A streaming approach is not that low-hanging, in my opinion. What we can do is to specifically look into the state change from offline to online, and kind of exceptionally allow larger online queues. (if enough RAM available).

---
:point_right: Thank you for supporting VerneMQ: https://github.com/sponsors/vernemq
:point_right: Using the binary VerneMQ packages commercially (.deb/.rpm/Docker) requires a paid subscription.

rtuin:
[...]

PR #2143, opened 2023-04-23
Allow ignore max online from offline queue to online queue

mths1:
## Proposed Changes

Currently, when a session goes from offline to online and the actual offlines messages > max_online_messages those additional messages are immediately dropped. 

This PR introduces:
(a) a way to switch between different strategies for moving messages from the offline queue to the online queue
(b) a new strategy called "add_all" which does the following

* it moves all messages (or to be more correct their metadata/references) from offline queue to the online queue of the sessions immediately
* it ignores the size limit of the online queue
* it utilizes the already existing streaming/batching functionality, that means the messages will not be loaded from disk immediately but within the batching from disk. The additional memory needed therefore is limited to the metadata.
* the offline message into the shadow "backup" queue which is used for batching allowing online messages to arrive 
* the order in which the messages are sent to the subscribers is not changed (offline before online)

it does not change the state machine or anything else, as it just "simulates" a online queue with a higher capacity.

## Types of Changes

- [X] New feature (non-breaking change which adds functionality) #1663 

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/apps/vmq_server/priv/vmq_server.schema
+++ b/apps/vmq_server/priv/vmq_server.schema
@@ -269,6 +269,13 @@
                                                                   {datatype, atom},
                                                                   hidden]}.
 
+%% @doc Allows a session that changes from offline to online to override the maximum
+%% online message count (max_online_messages). All offlines messages will be added 
+%% to the online queue. 
+{mapping, "override_max_online_messages", "vmq_server.override_max_online_messages", [{default, off},
+                                                  {datatype, flag}
+                                                  ]}.
+
 %% @doc specify how the queue should process the messages,
 %% either the 'fifo' or 'lifo' way.
 {mapping, "queue_type", "vmq_server.queue_type", [{default, fifo},
--- a/apps/vmq_server/src/vmq_queue.erl
+++ b/apps/vmq_server/src/vmq_queue.erl
@@ -79,6 +79,7 @@
     backup = queue:new(),
     type = fifo,
     max,
+    ignore_max = false,
     size = 0,
     drop = 0
 }).
@@ -227,6 +228,7 @@ default_opts() ->
 %%%===================================================================
 online({change_state, NewSessionState, SessionPid}, State) ->
     {next_state, online, change_session_state(NewSessionState, SessionPid, State)};
+%
 online({notify_recv, SessionPid}, #state{id = SId, sessions = Sessions} = State) ->
     #session{queue = #queue{backup = Backup} = Queue} = Session = maps:get(SessionPid, Sessions),
     cleanup_queue(SId, Backup),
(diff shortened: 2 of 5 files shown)


apps/vmq_server/priv/vmq_server.schema
```diff
@@ -269,6 +269,21 @@
                                                                   {datatype, atom},
                                                                   hidden]}.
 
+%% @doc specify how offline messages should be delivered if a session switches from offline
+%% to online. Currently, two strategies are supported: 'adhere_queue_size' or 'add_all'.
+{mapping, "queue_offline_online_strategy", "vmq_server.queue_offline_online_strategy", [{default, adhere_queue_size},
+                                                  {datatype, atom},
```
ioolkos, line 275: datatype can be a flag instead of enum, then you can just set a single setting to on or off. (`override_max_online` or sth similar)

apps/vmq_server/src/vmq_queue.erl
```diff
@@ -1060,15 +1084,27 @@ insert(MsgOrRef, #state{id = SId, deliver_mode = balance, sessions = Sessions} =
     State#state{sessions = maps:update(RandomPid, UpdatedSession, Sessions)}.
 
 session_insert(SId, #session{status = active, queue = Q} = Session, MsgOrRef) ->
-    {send(SId, Session#session{queue = queue_insert(false, MsgOrRef, Q, SId)}), MsgOrRef};
+    {
+        send(SId, Session#session{
+            queue_to_session_batch_size = 100, queue = queue_insert(false, MsgOrRef, Q, SId)
```
ioolkos, line 1089: hard coded batch size, wondering what to base this on. The session process can send a lot of messages to the consumer, depending on max_inflight/receive_max_client (in case QoS >1).

apps/vmq_server/priv/vmq_server.schema
mths1: Could do that. The idea was being able to add more sophisticated "strategies", if needed.

apps/vmq_server/src/vmq_queue.erl
mths1: Thanks. That is actualy a left-over from testing/development. The standard batch size of "100" was in there before. I will remove that line and it will revert to whatever queue_to_session_batch_size was.  

BatchSize = maps:get(queue_to_session_batch_size, Opts, 100),

apps/vmq_server/priv/vmq_server.schema
ioolkos: I think the current approach takes us a long way before more strategies are needed. I also think a flag just makes it easier for users, so I prefer that.

apps/vmq_server/src/vmq_queue.erl
```diff
@@ -786,6 +789,9 @@ add_session_(
         cleanup_on_disconnect := Clean
     } = Opts,
     BatchSize = maps:get(queue_to_session_batch_size, Opts, 100),
+    IgnoreMaxQueueSize =
```
ioolkos, line 792: What does MaybeIgnoreMaxMsgSize do and why the `and` here?

mths1: It is meant as a precaution that online the offline -> online transition can overwrite the max_online_messages and not any other usage of add_session_. If "MaybeIgnoreMaxMsgSize" it might ignore the size of the max online message queue. Thinking about it, the naming is odd because it could be misinterpreted as the maximum message size, while I meant the maximimum messages size of the queue. Let me know what you think of the switch and I will rename it to something better (if the naming caught your attention)

ioolkos: AllowExtendedQueueSize or similar...and I think here's one of the places where a short comment would be helpful in the code.

apps/vmq_server/priv/vmq_server.schema
ioolkos: Can you change the datatype to enum instead of atom?

ioolkos (approved):

PR #2153, opened 2023-05-09
Use default reg view for status page and flatten offline clients

mths1:
## Proposed Changes

Use default reg view for status page and flatten offline clients.

## Types of Changes

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/apps/vmq_server/priv/static/js/status.js
+++ b/apps/vmq_server/priv/static/js/status.js
@@ -109,6 +109,14 @@ $(function() {
         }
     }
 
+    function flatten_minus_one(val) {
+        if (val == -1) {
+            return 0;
+        } else {
+            return val;
+        }
+    }
+
     function cap_zero(val) {
         if (val < 0) {
             return 0;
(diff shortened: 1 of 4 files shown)


apps/vmq_server/priv/static/js/status.js
```diff
@@ -109,6 +109,14 @@ $(function() {
         }
     }
 
+    function flatten_minus_one(val) {
```
ioolkos, line 112: We might just use cap_zero then but why are we "off by one" here?

mths1: It is a timing issue. There is a short period of time when the total queues is smaller than the nr. of active mqtt connections. This can be seen when a lot of clients subscribe at once to an otherwise "clean" broker. The offline client then switches between -1 and 0 depending on the timing (and stays constantly at zero once all clients connected). I did not cap it at zero, because a -2..-n would point to a different problem. 

https://github.com/vernemq/vernemq/blob/7eab48515756114a573c3f6f621de7ba2e900318/apps/vmq_server/src/vmq_status_http.erl#LL85C9-L85C9

ioolkos: Heh, yes, a concurrency thing. The queue setup takes some time and is triggered by the connecting client session.
The hawk eyes you have!

ioolkos:
@mths1 let's consider this a bugfix, so still warrants adding a line to the changelog.

ioolkos (approved):

Issue #2184, opened 2023-08-14
extra info for packet_exceeds_max_size


nirfr:
### Describe your issue or feature request

Once in a while we get a lot of these messages:

session stopped abnormally due to 'packet_exceeds_max_size'

I guess its a faulty client, but I can't tell which...

### Describe the solution you'd like

Printing the client id with the error message would help tracking the origin/source of the problem

### Environment

- VerneMQ Version: vernemq/vernemq:1.12.3-alpine
- OS: k8s
- Erlang/OTP version (if building from source):
- Cluster size/standalone:

### Code of Conduct

- [X] I agree to follow the VerneMQ's Code of Conduct

ioolkos:
@nirfr Thanks, I see your point. We probably should get the session state here and get the subscriber_id for logging.
https://github.com/vernemq/vernemq/blob/de87dad923ad7a1cb26f4df842e0381ee4ead61c/apps/vmq_server/src/vmq_ranch.erl#L187

Right now, if you cannot conclude state information from one of the logs (console.log/error.log/crash.log) the client information is lost. You can maybe check for this in your log files, but I'll label this as an enhancement.

---
:point_right: Thank you for supporting VerneMQ: https://github.com/sponsors/vernemq
:point_right: Using the binary VerneMQ packages commercially (.deb/.rpm/Docker) requires a paid subscription.

mths1:
Here is a patch I wrote some time ago to fix this
https://github.com/mths1/vernemq/tree/feature/print-subscriber-on-log

unfortunately, I did not yet test it to the end...

PR #2186, opened 2023-08-15
Add additional logging information (print subscriber_id)

mths1:
## Proposed Changes

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #2184 )

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/apps/vmq_server/src/vmq_ranch.erl
+++ b/apps/vmq_server/src/vmq_ranch.erl
@@ -177,14 +177,15 @@ loop_({exit, Reason, State}) ->
     _ = internal_flush(State),
     teardown(State, Reason).
 
-teardown(#st{socket = Socket}, Reason) ->
+teardown(#st{socket = Socket, fsm_mod = FsmMod, fsm_state = FsmState}, Reason) ->
     case Reason of
         normal ->
             lager:debug("session normally stopped", []);
         shutdown ->
             lager:debug("session stopped due to shutdown", []);
         _ ->
-            lager:warning("session stopped abnormally due to '~p'", [Reason])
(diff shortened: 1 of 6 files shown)


apps/vmq_server/src/vmq_ranch.erl
```diff
@@ -177,14 +177,15 @@ loop_({exit, Reason, State}) ->
...
-teardown(#st{socket = Socket}, Reason) ->
+teardown(#st{socket = Socket, fsm_mod = FsmMod, fsm_state = FsmState}, Reason) ->
     case Reason of
         normal ->
             lager:debug("session normally stopped", []);
         shutdown ->
             lager:debug("session stopped due to shutdown", []);
         _ ->
-            lager:warning("session stopped abnormally due to '~p'", [Reason])
+            SubscriberId = apply(FsmMod, subscriber, [FsmState]),
```
ioolkos, line 187: FsmState should already carry the subscriber id here. Am I missing a point? (reason for the function call)

mths1: That is true, but there are three different state records from the three fsms that could appear here (mqtt, mqtt5, mqtt_pre), so the actual record used depends on the FsmMod. I am still an Erlang beginner, but it seems Erlang does not allow one to reference a record (or export it's definition) that is defined in another module. I found only two ways of implementing it, one the function call (which is called only in an error case, so it should be ok performance wise) or moving the record definition into header files and reference them directly. Keeping the record module-locale and encapsulate it via a function seemed more natural to me. Let me know if you prefer the header idea, or if there is a third way I do not know about yet.

mths1:
@ioolkos : this one should be good to go as well

PR #2197, opened 2023-09-20
Feature database client id match

bilal-haider-cowlar:
## Proposed Changes

As we were using username and password in acl file auth. But in the case of DB, we are also matching client_id. There are a lot of cases where we only want to authenticate using username and password from the DB. So, i made this thing optional. 

Now we only have to pass a variable in our docker env like  `DATABASE_CLIENT_ID_MATCH: 'off'`. If its value is off then client_id will not get matched from DB and it will act same like in case of acl files auth. 

I just access this variable from os env and i have changed swl query for not matching client_id.

## Types of Changes

## Checklist

## Further Comments

For now. I have implemented this optional feature for mysql. I want to get reviews on it. If this get approved then I will add this functionality for all databases.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -23,7 +23,7 @@ jobs:
         run: |
           sudo apt-get install -y libsnappy-dev
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Cache PLTs
         id: cache-plts
         uses: actions/cache@v3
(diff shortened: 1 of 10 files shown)


bilal-haider-cowlar:
@ioolkos want reviews on this

ioolkos:
@bilal-haider-cowlar Thanks for the effort. THB, I'm not yet sure what to do, as it introduces an option to lower security (but only for MySQL). Also note that if we deem this acceptable we should have the setting in the vmq_diversity schema file, and not have it read from the OS user env.

mths1:
I can see the need (acl based on "device class") without having to have an acl on each individual device.  If we want to proceed with this pull request I'd suggest to make it even more flexible, as in just allow a null in the db and do an auth like (pseudo sql)

select top 1 ... where ... and ((client_id := client_id) or (client_id is null)) order by client_id nulls last

other options would include auth based on device classes (which would be an additional concept) or allow some sort of regex...

Climax777:
This is also an issue for me using the MongoDB method

PR #2219, opened 2023-11-22
Add a new delayed subscription strategy for topics with retained mess…

mths1:
Add a new delayed subscription strategy for topics with retained messages. Due to the fact that the retrain service and the message push are separate subsystems this is mitigate the unfortunate situation when a subscriber subscribes slightly after the messaging system has send the message, and slightly before the retrain sync has completed. This only affects clustered setups which use retrained messages.

## Proposed Changes
- Add a new strategy that let subscribers wait for the retrain subystem to sync
- New messages will still be accepted, just delayed until the subsystem has synced
- After the sync, the retrained message is added to the beginning of the queue

Implications:
- The new function QFun in vmq_queue might hurt performance slightly - did not measure yet. Could maybe rewritten in a way that it uses pattern matching... 
- It is possible to get the very same message twice (the retrained one, and the original one, based on some unfortunate timing)

This PR is just to see if there is interest in finalizing this, or if we want to stick with the original design. 

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #1633 )

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/apps/vmq_server/priv/vmq_server.schema
+++ b/apps/vmq_server/priv/vmq_server.schema
@@ -43,6 +43,23 @@
  [{default, on},
   {datatype, flag}]}.
 
+%% Retrain subscriber mode defines if subscribers on topics with retrained flag set should wait for 
+%% a retrain server sync or process immmediatly.
+{mapping, "subscriber_retain_mode", "vmq_server.subscriber_retain_mode", [
+                                                               {default, immediate},
+                                                               {datatype, atom},
+                                                               hidden
+                                                              ]}.
+
+{translation, "vmq_server.subscriber_retain_mode",
+ fun(Conf) ->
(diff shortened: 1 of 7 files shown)


rafaelnobrekz:
Thanks @mths1 for the proposal. Is there an easy way to test your PR on a server using VerneMQ?
@ioolkos Is there interest in adopting this kind of solution to the main code? Or is there another solution being discussed elsewhere? In our use case, there is initially no retained messages, so when we subscribe close to the latest publish, we get no messages at all and have to manually resubscribe after some threshold to get the published retained message.

ioolkos:
@rafaelnobrekz you can test this PR by:

```
git fetch origin pull/2219/head:2219
git checkout 2219
```
You then have to compile this branch.
This PR is not a full solution but might mitigate the issue. Thank you for testing!

PR #2228, opened 2023-12-06
Optional feature for forced disconnect on unauthorized mqtt31 publish

MeisterMasi:
## Proposed Changes

For MQTT 3.1 there is a gap in the definition what do if authorization of a publish request fails, only the happy path is described and defines to answer with acknowledgements depending of the QoS level.
So for MQTT 3.1 clients with no possibility to update to MQTT 3.1.1 there is no way to figure out, if the message was accepted by the authentication or not. Only breaking MQTT 3.1 compatibility and forcing a disconnect may help to prevent sending the ACK to the client in case of an authentication failure. 

## Types of Changes

## Checklist

### DIFF
--- a/apps/vmq_server/src/vmq_mqtt_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt_fsm.erl
@@ -81,6 +81,9 @@
     %% present and default value if not present.
     def_opts :: map(),
 
+    %% disconnect on unauthorized publish, even for non MQTT 3.1.1 clients
+    disconnect_on_unauthorized_publish_v3 = false :: boolean(),
+
     %% TODO
     trace_fun :: undefined | any()
 }).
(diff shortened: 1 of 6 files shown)


ioolkos:
Discussion.
MQTT 3.1.1 spec (but not 3.1) says:
"If a Server implementation does not authorize a PUBLISH to be performed by a Client; it has no way of informing that Client. It MUST either make a positive acknowledgement, according to the normal QoS rules, or close the Network Connection [MQTT-3.3.5-2]."

In Verne, we close the connection if a PUBLISH is not authorised. It just occurred to me that the decision on what strategy to use could be left to the user by introducing a config switch.
I'd be fine using the same decision for both 3.1 and 3.1.1.

OTOH, I guess there are no use cases that are critically dependent on staying silent on non-authorized PUBLISHs for 3.1.

If anyone has an opinion on this, please share.

mths1:
My only comment is that it would change existing behaviour to have a single switch for both 3.1 and 3.1.1, otherwise I'd support to have this configurable (but I do not have 3.1 clients, so the problem does not apply to me :-) )

apps/vmq_server/src/vmq_mqtt_fsm.erl
```diff
@@ -1034,6 +1042,10 @@ dispatch_publish_qos0(_MessageId, Msg, State) ->
             %% we have to close connection for 3.1.1
             _ = vmq_metrics:incr_mqtt_error_auth_publish(),
             {error, not_allowed};
+        {error, not_allowed} when DisconnectOnUnauthorizedPublishV3 ->
```
ioolkos, line 1045: Note we could use a single clause instead of 2, using a guard sequence:
`{error, not_allowed} when ?IS_PROTO_4(Proto) ; DisconnectOnUnauthorizedPublishV3 ->`   ... etc

ioolkos (approved):

PR #2248, opened 2024-02-12
Add simple options to HTTP health endpoint

ioolkos:
Some options in addition to `/health`:
EDIT:
changed paths to
`/health/ping`: Cowboy (ie. Verne) is up.
`/health`: behaviour as before
`/health/listeners`: will fail if any of the configured listeners is down or suspended
`/health/listeners_full_cluster`: will fail if any listener is down or any of the cluster nodes is offline. (you probably don't want to use this to base automated actions on the status)


mths1:
Wouldn't a "path extension" be more natural? as in health/ping ? health/listerners?

ioolkos:
With HTTP I'm never sure what's "natural" :)
But this seems like a good idea. But it doesn't work for combination, or does it?

PR #2269, opened 2024-03-16
Enable the Bridge client to connect to Inet6 MQTT endpoints

ioolkos:
Doc:
Introduces new vmq_bridge config settings:

`vmq_bridge.tcp.default.inet_version = inet|inet6`
`vmq_bridge.ssl.default.inet_version = inet|inet6`

with `default` being an arbitrary bridge name, informing `default` that the other bridge endpoint to contact is an IPv4 or IPv6 endpoint.

### DIFF
--- a/apps/vmq_bridge/priv/vmq_bridge.schema
+++ b/apps/vmq_bridge/priv/vmq_bridge.schema
@@ -267,6 +267,13 @@
                                                                     {include_default, "br0"},
                                                                     {commented, 10}
                                                                    ]}.
+%% @doc For outgoing bridges, let us know whether the other endpoint is
+%% 'inet' or 'inet6' (an IPv4 or an IPv6 address)
+{mapping, "vmq_bridge.tcp.$name.inet_version", "vmq_bridge.config", [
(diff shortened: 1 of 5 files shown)


mrceyhun:
Hi @ioolkos , many thanks for this PR. Are you planning to make any pre-release or RC including this changes? It would be great to have.

mrceyhun:
One more question: if a new release will take time, do you think we can cherry-pick this commit to a release `1.12.4` ( or `1.13`) without having any issue?

ioolkos:
@mrceyhun this will be part of the final release of 2.0, yes. It'd be great if someone can test the feature with their use case.

mrceyhun:
Thanks @ioolkos . Do you think it is ok to cherry-pick this commit to the `1.13`(or `1.12.4`) release? If so, we can test and report our use-case of this PR.

ioolkos:
@mrceyhun Maybe, but I cannot review the commit history right now to *guarantee* it to you.

mrceyhun:
Got it, thank you!

PR #2271, opened 2024-03-16
Update default rule for all clients

ashtonian:
merge password obfuscation with default rule and closes https://github.com/vernemq/vernemq/pull/2054


ioolkos:
I was looking for a mitigation for the following aspect:
With the default rule enabled, a client knowing any valid username/password combo can connect an arbitrary number of clients (and they will stay connected, even if the ACL would not allow any writing or reading from topics).
I don't see an immediate technical approach to avoid this. 

If we merge this, we should make users aware of that consequence when adding a default rule.

ioolkos:
@ashtonian did you do functional testing on all adapted database scripts, or just the one DB you use?
Adding some quick tests would be possible, but I don't think of it as a blocker.

ashtonian:
> I was looking for a mitigation for the following aspect: With the default rule enabled, an MQTT client knowing any valid username/password combo can connect an arbitrary number of clients (and they will stay connected, even if the ACL would not allow any writing or reading from topics). I don't see an immediate technical approach to avoid this.


Just want to make sure I understand the issue here - with a clientid:*, user, pass combo enables an unlimited number of clients to connect? But if a user tried to match with a different clientid, user, pass combo they would still only be limited to 1 session since it doesn't match the user/pass.

One way to change that would be to add a separate field to limit # of connections per auth entry with a default value of 1, but you would then also need to track/update the active number somewhere. 

I have only tested the pg one.

ioolkos:
@ashtonian cool, thanks. Okay, hehe, it seems I misremembered what the intention was here. You still enforce alignement of ClientId, Username, PW, so there's no risk of unlimited ClientIDs.
Man, I should have coffee before being on Github :)

ashtonian:
Haha, totally understand, interacting with coding before coffee is dangerous.

ioolkos (approved):

PR #2293, opened 2024-05-19
Initial support for compile with OTP-27

mths1:
## Proposed Changes

Initial support for compile with OTP-27. 
Notes:
* Nodetool works only in interpreted mode. OTP27 by default, compiles escripts. Unfortuntely, vernemq stop somehow crashes in compile mode. In interpreted mode, which was the default in OTP26 and earlier, it works as expected. We need to fix this for OTP28.
* code:lib_dir/2 has been deprecated. Needs to be solved in OTP29.

Things to come:
* JSON module support (follow up pull request). OTP27 does not need jsx anymore, and vmq_json can support both for a while and drop jsx with OTP29.

## Types of Changes

- [x] DevOps (Build scripts, pipelines...)

## Checklist

## Further Comments

considered, etc.


.github/workflows/pr.yml
```diff
@@ -161,10 +161,13 @@ jobs:
           # Monterrey
           - macos-12
```
ioolkos, line 162: Apparently macos-13 and macos-14 are available as Github runners. Should we switch to a newer macos here anyway?

rebar.config
```diff
@@ -68,6 +68,8 @@
         vmq_server,
         vernemq_dev,
         credentials_obfuscation,
+        wx,
```
ioolkos, line 71: For a pure server build, we don't need `wx` but we could use multiple build profiles and have a profile for developing.

mths1: Oh, yeah. That is a left-over from debugging. Adding a profile which supports the debugger would be nice. Once I got it running, I have fallen in love with the gui debugger ;-)

ioolkos: it is cool!
to add a profile, we can add sth like the following to the Makefile:
```
dev_build: PROFILE = as develop
dev_build: rel
```

Then add `profiles` to the rebar.config (keeping the main relx build out of it):

```
{profiles, [
   {develop, [
        {relx, [
            {release, {vernemq, semver}, [
                sasl,
                inets,
                xmerl....]}]}]}
                
           ]}.
```
then build with     
`make dev_build`

PR #2325, opened 2024-06-26
Add fold fun to delete expired retained messages

ioolkos:
If a retained msg has an expiry, the check to delete it is currently triggered by the actual delivery.
This PR adds a fold to check for and delete residual (expired) messages from the retain cache. Consider this a workaround; we should fully automate expired msg deletion, covering the stated edge case as well.

Performance for big cache sizes untested.

### DIFF
--- a/apps/vmq_server/src/vmq_info_cli.erl
+++ b/apps/vmq_server/src/vmq_info_cli.erl
@@ -27,6 +27,7 @@ register_cli() ->
     vmq_session_reauthorize_cmd(),
     vmq_retain_show_cmd(),
     vmq_retain_delete_cmd(),
+    vmq_retain_delete_expired_cmd(),
     vmq_loq_show_cmd(),
     vmq_log_level_cmd(),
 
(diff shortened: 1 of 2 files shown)


ioolkos:
If you just want to add this function (ie not hot loading the full module), for instance in older releases, you can do the following:

1. Create a file called `vmq_retain_exp.erl`:

```erlang
-module(vmq_retain_exp).
-export([fold_expired/0]).

fold_expired() ->
    Fun = fun
        ({{MP, Topic}, {retain_msg, _, #{p_message_expiry_interval := _}, {TS, _}}}, Acc) ->
            case vmq_time:is_past(TS) of
                true ->
                    vmq_retain_srv:delete(MP, Topic),
                    Acc + 1;
                _ ->
                    Acc
            end;
        ({{_, _}, {retain_msg, _, _, _}}, Acc) ->
            Acc
    end,
    ets:foldl(Fun, 0, vmq_retain_srv).
```

2. Copy the file to `erlio-patches`:
```
sudo cp vmq_retain_exp.erl /usr/lib/vernemq/lib/erlio-patches 
```
3. Attach to vernemq and compile the file:
```
sudo vernemq attach
```
```
(VerneMQ@127.0.0.1)1> c("lib/erlio-patches/vmq_retain_exp.erl").
{ok,vmq_retain_exp}
```
You only have to compile this once. VerneMQ will load the files from erlio-patches after booting (so we compile the file here to avoid rebooting VerneMQ).

4. Call our function to fold for expired retain messages:
```
(VerneMQ@127.0.0.1)2> vmq_retain_exp:fold_expired().
0
```
5. Safely detach from VerneMQ, by tying `ctrl+g` (that's a `g` like George), then at the user switch type `q`

To call the function on a system console (without attaching to VerneMQ), you can use a line like this, adapting the VerneMQ node name and the cookie (`-c`) to your case:

```
 erl_call -name 'VerneMQ@127.0.0.1' -c vmq -a 'vmq_retain_exp fold_expired []'
```

In a cluster, it should be sufficient to do this on a single node, as the deletion of retained messages should propagate to the other nodes. Make sure to test, though.

PR #2347, opened 2024-09-01
vmq_swc: Use separate DB for DKM store

ioolkos:
This PR singles out the DKM store to separate it from storing Objects.

This, again, is an experiment to speed up node sync. It is not compatible with existing on-disk metadata stores.
To reset a node `data/swc_meta` and `data/swc_dkm` have to be deleted now.


marcusramberg:
Been testing this a bit now. Looked very promising after replacing one pod at a time twice, without ending in empty sync, but after a rolling restart the graphs now look like this:

image

Went into the pods to confirm that they all started with empty data folders.

marcusramberg:
Did an initial build of the latest commit here, and did a rollout restart after deploying on a cluster that was already in empty sync, and it seems to have cleared it, so very promising! Will do some more testing under load tho.

ioolkos:
@marcusramberg Thanks again for testing.
This is not yet in a final state. I want to add an initial fast sync, and do some procedural testing under lab conditions. For instance, doing a re-join, let the node sync up, and then do 1 additional simple stop/start of that leaving/joining node before rolling any others allows me a full rolling restart with my current version.

ioolkos:
@marcusramberg feel free to test the latest commit.

It's an idea (POC) that should fix:
- all empty sync
- issues with rolling re-joins
- issues with increased CPU/RAM after joins

Todo:
- [ ] Adapt tests
- [ ] Test against various load scenarios
- [ ] Test Retain store
- [ ] Maybe add fast initial sync

marcusramberg:
Rebuilt this branch and did some testing. Seems I'm getting a lot of crashes after a rolling restart, along with empty sync. Here's a sample from the pod logs:

https://gist.github.com/marcusramberg/91f0c3cc9d3c89c0ccb9b41a38ad87f8

ioolkos:
@marcusramberg Thanks a lot for testing. Something somehow is still different in our setups :grimacing: 
I don't fully understand your test scenario, but I'll setup something that does a ramp-up with constant connect/disconnects. I think that's what you do.
I also suppose I should start test this on Kubernetes then.

ioolkos:
A `rollout restart statefulset` weirdly shows inconsistent behaviour. I see some nodes leave the cluster, some nodes not.
Deleting the pods one by one manually, consistently makes the nodes leave and re-join the cluster.

Issue #2355, opened 2024-10-06
[Bug]: NULL char in CONNECT packets


tiz314:
### Environment

- VerneMQ Version: 1.13.0
- OS: Debian 6.1.112-1 x86_64
- Cluster size/standalone: standalone

### Current Behavior

When sending a CONNECT packet, the broker ACKs the connection even if the clientID contains a NULL char. The client is then able to perform any expected operation like a client connected using a valid clientID. 

At the moment, no related vulnerabilities are known. 

### Expected behaviour

As stated at line 268, 274 and 3474 of the [MQTTv5 standard](https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.pdf) if a packets contains an UTF-8 string that presents a NULL char it must be considered as a malformed packet, and if the broker receives a malformed packet it must close the connection. 

### Code of Conduct

- [X] I agree to follow the VerneMQ's Code of Conduct

ioolkos:
@tiz314 Thanks, yes, we need to add the NULL check to the `ensure_utf8` functions in both parsers. Do you want to have a go at it? ;)

---
👉 Thank you for supporting VerneMQ: https://github.com/sponsors/vernemq
👉 Using the binary VerneMQ packages commercially (.deb/.rpm/Docker) requires a paid subscription.

tiz314:
Sure! If you want I'll make a PR :)

PR #2356, opened 2024-10-08
Add NULL check in ensure_utf8

tiz314:
## Proposed Changes

Added check to verify that any utf8 string doesn't contain a NULL character, which is not allowed by the MQTT standard. This PR fixes the issue #2355.

This change is applied to both `ensure_utf8` functions in `vmq_parser.erl` and `vmq_parser_mqtt5.erl`

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #2355 )

## Checklist


ioolkos:
@tiz314 Thanks!
I wonder whether we want to add a test case for this, where we'd generate a Connect packet with something like `gen_connect(<<0>>, [])`.
Also, I suppose NULL is forbidden for the MQTT username and any other UTF-8 strings here we use `ensure_utf8` as well), but I haven't yet looked into that.

tiz314:
Of course! Following the [standard](https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.pdf) it says that no UTF-8 encoded string are allowed to contain U+0000, so we should check every field in every possible MQTT packet type and add a check on them if the standard says it is processed as a string. 

As regards the test case, if you give me a go I can add a connect test in `vmq_server/test`

ioolkos:
Adding a test to the vmq_server connect suite is great!
A previous commit by @mths1 added the UTF-8 verification for Username and Topic as well.

tiz314:
I'm quite struggling with the development of the test case. Do you use common test? Which tool do you use to run it? I tried with rebar3 but it gives me errors about files not found, such as:

```
vmq_commons/src/vmq_types_common.hrl:{3,14}: can't find include lib "vernemq_dev/include/vernemq_dev.hrl"; Make sure vernemq_dev is in your app file's 'applications' list

vmq_commons/src/vmq_types_mqtt.hrl:{1,14}: can't find include lib "vernemq_dev/include/vernemq_dev.hrl"; Make sure vernemq_dev is in your app file's 'applications' list

vmq_commons/src/vmq_types_mqtt5.hrl:{1,14}: can't find include lib "vernemq_dev/include/vernemq_dev.hrl"; Make sure vernemq_dev is in your app file's 'applications' list
```

ioolkos:
@tiz314 Here's how to proceed.
- In the vernemq dir, run `make rel` so that you have a default build (this will create `default` build in the `_build dir`)
- Run the full common test suites with `rebar3 ct`
- Alternatively, run a single test suite only with something like  `rebar3 ct --suite apps/vmq_webhooks/test/vmq_webhooks_SUITE.erl` (if you want, you can even just run single test cases)

Running any command with `rebar3 ct` will create a `test` build in the `_build` dir. In the `test` dir, you'll also have logs for your test runs. You'll also see HTTP files which you can open in a browser.

tiz314:
I pushed a first version of the test case but I'm having some serious difficulties in catching the patched parser exception that gets raised when sending the null char, making the test always pass. 

First time with this kind of tech :(

ioolkos:
@tiz314 Thanks for your effort, much appreciated. I'll look into the test case.

PR #2374, opened 2025-01-16
Init feature: forwarding connection opts (SSL client cert) to auth pl…

ioolkos:
WIP

This adds a settable flag for SSL listeners:

`listener.ssl.my_listener.forward_connection_opts = on`

If on, this will forward connection info from the TLS layer to your authentication plugin. Currently, this is used to forward the full TLS client cert, so that your authn/authz plugin can use information from the client cert.

Cannot be used with Lua plugins atm.

Example:
Configure an SSL listener, with `forward_connection_opts` and `require_certificate` set to `on`:

```
listener.ssl.test = 127.0.0.1:8885
listener.ssl.test.cafile = ./ssl/all-ca.crt
listener.ssl.test.keyfile = ./ssl/server.key
listener.ssl.test.certfile = ./ssl/server.crt
listener.ssl.test.require_certificate = on
listener.ssl.test.forward_connection_opts = on
```
Webhooks example (Python hook). You can now grab the 64encoded client cert from `data`:

```python
def handle_auth_on_register(data):
    # Cache example
    web.header('cache-control', 'max-age=30')
    # only allow user 'joe' with password 'secret', reject all others.
    print ('cert:', data['client_cert'])
    if "joe" == data['username']:
        if "secret" == data['password']:
            return json.dumps({'result': 'ok'})  
    return json.dumps({'result': {'error': 'not allowed'}})
```

### DIFF
--- a/apps/vmq_server/priv/vmq_server.schema
+++ b/apps/vmq_server/priv/vmq_server.schema
@@ -1826,6 +1826,37 @@
                                                                                   hidden
                                                                                  ]}.
 
+%% @doc If 'listener.ssl.forward_connection_opts' is enabled, VerneMQ will
+%% forward the full client certificate to the authentication plugin (in the
+%% auth_on_register hooks).
+%% Configure on protocol level (for all listeners):
+%%
+%%     - listener.ssl.forward_connection_opts
+%%     - listener.wss.forward_connection_opts
+%%
+%% or on the listener level:
+%%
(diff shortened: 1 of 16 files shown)


ioolkos:
Note: we could have `forward_connection_opts` for TCP listeners also. It would obviously not forward TLS info, but it could tell a plugin through which listener a request is coming in. The latter use case would currently require mountpoint'ed listeners.

Issue #2364, opened 2024-10-29
[Bug]: vmq_bridge exited with reason: shutdown


coffeeheart:
### Environment

- VerneMQ Version: 1.12.3
- OS: openEuler 20.03
- Erlang/OTP version (if building from source): 23.0
- Cluster size/standalone: standalone

### Current Behavior

Hello!
There are 38 independent nodes. One center node sets up 37 bridges with other 37 nodes.
One of the 37 nodes cut off the electricity for about 20 minutes, then the center node's vmq-bridge plugin exited. And the vernemq process stayed running.
[crash.log](https://github.com/user-attachments/files/17554458/crash.log)
[error.log](https://github.com/user-attachments/files/17554460/error.log)

### Expected behaviour

The vmq-bridge plugin doesn't exit. Or it can restart automatically.

### Configuration, logs, error output, etc.

```markdown
vmq_bridge.tcp.br0 = 50.15.180.37:1882
vmq_bridge.tcp.br0.client_id = core_vernemq_br0
vmq_bridge.tcp.br0.cleansession = off 
vmq_bridge.tcp.br0.topic.1 = /radar/room/backup_state/*  out 1
vmq_bridge.tcp.br0.topic.2 = /meeting_core/meetings/*  in
vmq_bridge.tcp.br0.username = ""
vmq_bridge.tcp.br0.password = ""
vmq_bridge.tcp.br1 = 50.15.179.197:1882
vmq_bridge.tcp.br1.client_id = core_vernemq_br1
vmq_bridge.tcp.br1.cleansession = off 
vmq_bridge.tcp.br1.topic.1 = /radar/room/backup_state/*  out 1
vmq_bridge.tcp.br1.topic.2 = /meeting_core/meetings/*  in
vmq_bridge.tcp.br1.username = ""
vmq_bridge.tcp.br1.password = ""
vmq_bridge.tcp.br2 = 50.15.178.101:1882
vmq_bridge.tcp.br2.client_id = core_vernemq_br2
vmq_bridge.tcp.br2.cleansession = off 
vmq_bridge.tcp.br2.topic.1 = /radar/room/backup_state/*  out 1
vmq_bridge.tcp.br2.topic.2 = /meeting_core/meetings/*  in
vmq_bridge.tcp.br2.username = ""
vmq_bridge.tcp.br2.password = ""
vmq_bridge.tcp.br3 = 50.15.74.70:1882
vmq_bridge.tcp.br3.client_id = core_vernemq_br3
[...]

PR #2379, opened 2025-02-06
vmq_bridge: Handle and log unknown internal messages in bridge process

ioolkos:
Reason: We've seen unexpected events like tcp_closed end up in the bridge


penguinol:
Does this pr fix https://github.com/vernemq/vernemq/issues/2364 ？

Issue #2406, opened 2025-05-21
bridge to NanoMQ causes a loop of messages


alessiopollero:
### Environment

- VerneMQ Version: 2.0.1
- OS: Docker image vernemq/vernemq:latest
- Cluster size/standalone: one node 

My docker compose file is similar to the following: 

```
networks:
  iot_internal_network:
    driver: bridge

services:
  vernemq:
    image: vernemq/vernemq:latest
    container_name: vernemq_broker
    networks:
      - iot_internal_network
    environment:
      - DOCKER_VERNEMQ_ACCEPT_EULA=yes
      - DOCKER_VERNEMQ_ALLOW_ANONYMOUS=on
      - DOCKER_VERNEMQ_PLUGINS.vmq_bridge=on
      - DOCKER_VERNEMQ_VMQ_BRIDGE.tcp.nanomqdatanet=172.90.90.1:1883
      - DOCKER_VERNEMQ_VMQ_BRIDGE.tcp.nanomqdatanet.cleansession=on
      - DOCKER_VERNEMQ_VMQ_BRIDGE.tcp.nanomqdatanet.client_id=test_bridge
      - DOCKER_VERNEMQ_VMQ_BRIDGE.tcp.nanomqdatanet.username=mqtt-user
      - DOCKER_VERNEMQ_VMQ_BRIDGE.tcp.nanomqdatanet.password=test
      - DOCKER_VERNEMQ_VMQ_BRIDGE.tcp.nanomqdatanet.topic.1=TTT/SLT9/* both 1
      - DOCKER_VERNEMQ_VMQ_BRIDGE.tcp.nanomqdatanet.restart_timeout=10
      - DOCKER_VERNEMQ_VMQ_BRIDGE.tcp.nanomqdatanet.try_private=off
    ports:
      - "1883:1883"



```

Whenever I send a message to to the topic `TTT/SLT9/CMD/SOMECOMMAND`
it looks like I run in a loop and messages and responses are entering in a loop basically generating so many messages that are going to overwhelm the NanoMQ instance on the other side of the bridge.
Is there a way to prevent this issue either built in or with some forwarding rules on the bridge configuration ? 

I don't know if this can be considered a bug or not, but probably as the VerneMQ should prevent this loop of messages in the first place I believe. 

Thanks in advance for any help.

### Current Behavior

Messages are looping on the topics between the two bridged MQTT brokers and causing resources exhaustion 

### Expected behaviour

[...]

PR #2408, opened 2025-05-22
Attempt to fix prefix parsing

aleroot:
## Proposed Changes

This pull request addresses a bug in the `vmq_bridge:remove_prefix/2` function within `apps/vmq_bridge/src/vmq_bridge.erl`. The previous implementation could lead to `function_clause` errors when a configured topic prefix (after being parsed into a list of segments) did not match the beginning of the actual topic it was intended to be stripped from. This occurred because there was no clause to gracefully handle such mismatches. Additionally, some of the existing recursive clauses for list-based prefixes were problematic and could lead to incorrect behavior.

The fix refactors `remove_prefix/2` by:
- Removing the faulty recursive clauses for list arguments.
- Introducing a new private helper function, `remove_prefix_recursive_helper/3`, which implements robust recursive logic for removing prefixes. This helper correctly compares segments and, in case of a mismatch or if the topic is exhausted before the prefix, it safely returns the original topic segments.
- Modifying `remove_prefix/2` to delegate to this new helper function for list-of-segments prefixes, while retaining its correct base cases for `undefined` and empty list `[]` prefixes.

These changes ensure that prefix removal during MQTT message bridging is more robust and reliable, preventing unexpected errors and ensuring correct topic manipulation as defined by the bridge configuration. This improves the stability of the bridge plugin, particularly for configurations using explicit local and/or remote prefixes for topic mapping, which was related to issue #2406.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #2406)

## Checklist

## Further Comments

The solution chosen involves introducing a new, more explicit helper function to handle the recursive logic for prefix removal.


ioolkos:
@aleroot Can you please run `rebar3 fmt` and add the format changes as a commit?

aleroot:
> @aleroot Can you please run `rebar3 fmt` and add the format changes as a commit?

Ah, sorry. Done, and updated PR.

PR #2427, opened 2025-08-06
fix: active_mqtt_connections to sum all active WebSocket connections …

VivekPipaliya23:
## Proposed Changes
This change will fix the issue where web socket metrics are not incrementing because the ranch process does not recognize WS connections as active MQTT connections and instead categorizes them under all connections.

image

image

allow_anonymous = on
listener.tcp.default = 127.0.0.1:1883
listener.ws.default = 127.0.0.1:9001

**Different Protocol Handlers:**
Direct MQTT connections (mqtt/mqtts) use the vmq_ranch protocol handler
WebSocket MQTT connections (mqttws/mqttwss) use the cowboy_clear protocol handler
Connection Counting Discrepancy:
Direct MQTT: Ranch correctly tracks active_connections for direct MQTT connections because each TCP connection maps directly to one MQTT session
WebSocket MQTT: Ranch tracks the TCP connection, but when the connection upgrades to WebSocket and then establishes an MQTT session, the active_connections count may not accurately reflect active MQTT sessions
Root Cause: The active_mqtt_connections() function in vmq_ranch_sup.erl was using Ranch's active_connections field directly for all listener types, but WebSocket connections have a different lifecycle where:
A TCP connection is established (counted in all_connections)

## Types of Changes

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,5 @@
+- Bugfix: active connections count for WS in metrics and listener info
+
 ## VerneMQ 2.1.1
 
 - Bugfix syslog boot and add log.syslog.level setting #2413


ioolkos:
@VivekPipaliya23 Thanks, so it looks we cannot offer an active connection count for WebSockets?

ioolkos:
@VivekPipaliya23 got you. I think we should fix this from a user perspective, not from the perspective what `all_connections` and `active_connections` are in Ranch.

With this PR, you can have one WS connection on the broker, but you'd still see:

```
gauge.active_mqtt_connections = 0
gauge.active_mqttws_connections = 0
gauge.total_active_connections = 0
```
Users would expect:

```
gauge.active_mqtt_connections = 0
gauge.active_mqttws_connections = 1
gauge.total_active_connections = 1
```

VivekPipaliya23:
@ioolkos, The above metric screenshot that I shared is after my change, where we can see it increments the WS count.
CMIIW.

PR #2484, opened 2026-02-28
cluster auto balancing

ashtonian:
## Proposed Changes

This is an attempt at a fairly simple in cluster auto-balancing solution with the goal of balancing device connections accross the cluster. Wanted to start a discussion because its a big feature set. This is is the initial piece. 

Strategy:
* Expose a http /balance endpoint that load balancers can query to make informed routing decisions (ie k8s readiness or NLB health check). These endpoints will return 503 when a node is overloaded/out of balance with the goal of allowing the load balancer to route traffic elsewhere. 
* Rejection hook - a hook that will hard reject connections on an over loaded node. This accounts for delay in lb action. 
* Active rebalance - this can run on an interval or be triggered via a command to smartly (don't disconnect the same device) disconnect clients from overloaded nodes. This will hopefully force them to reconnect to a healthy node. 

This introduces `vmq_balance_srv` that enables a node to periodically evaluate its own load based on active connection count. All opt in/ confgiruable. This PR is just the first part of the strategy. 

## Types of Changes

## Checklist

## Further Comments

This needs more testing and feedback.

### DIFF
--- a/apps/vmq_server/priv/vmq_server.schema
+++ b/apps/vmq_server/priv/vmq_server.schema
@@ -341,7 +341,7 @@
 %% @doc specifies the modules to be enabled by the http(s) listeners. Overriding this setting is possible for specific listeners.
 {mapping, "http_modules", "vmq_server.http_modules", [
                                                       {default, "[vmq_metrics_http,vmq_http_mgmt_api, vmq_status_http, vmq_health_http,
-                                                                vmq_http_v2_api]"},
+                                                                vmq_http_v2_api, vmq_balance_http]"},
                                                       {datatype, string},
                                                       hidden
                                                      ]}.
@@ -393,7 +393,7 @@
   fun(Conf) ->
      HTTPModules = cuttlefish_variable:filter_by_prefix("http_module",Conf),
      HTTPModulesAsAtom = lists:map(fun(Element) -> list_to_atom(lists:nth(2,element(1, Element))) end, HTTPModules),
-     KnownHttpModules = [vmq_metrics_http,vmq_http_mgmt_api, vmq_status_http, vmq_health_http, vmq_http_v2_api, vmq_http_pub, vmq_web_ui],
+     KnownHttpModules = [vmq_metrics_http,vmq_http_mgmt_api, vmq_status_http, vmq_health_http, vmq_http_v2_api, vmq_http_pub, vmq_web_ui, vmq_balance_http],
      UniqueModules = lists:usort(HTTPModulesAsAtom ++ KnownHttpModules),
      Map = lists:map(fun(Atom) -> String = "http_module."++atom_to_list(Atom)++".auth.mode", #{Atom => cuttlefish:conf_get(String, Conf, "noauth")} end, UniqueModules),
      Ret = maps:from_list(lists:foldl(fun(Elem, Acc) -> maps:to_list(Elem) ++ Acc end, [], Map)),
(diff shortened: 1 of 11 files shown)


ashtonian:
@ioolkos  - This is ready for deeper review and testing. Phase 1 - just exposes a health check endpoint to better inform l4 load balancers. 

# Cluster auto-balance (Layer 1)

## Summary

Adds an opt-in readiness signal that lets external load balancers steer new connections away from overloaded nodes. A periodic check compares this node's connection count against the cluster average; when the local node exceeds a configurable threshold, the readiness endpoint returns 503 until the imbalance subsides. Disabled by default, so existing deployments see no behavior change.

Intended primarily as a Kubernetes readiness probe or NLB target health check.

## Endpoint

### `GET /api/balance-health`

Returns the node's current readiness state with respect to cluster connection balance.

**Status codes**

| Code | Meaning | When |
|------|---------|------|
| `200` | `accepting` | Balance disabled, OR enabled and this node is at/below the threshold. |
| `503` | `rejecting` | Balance enabled and this node is above `cluster_avg * threshold`. |

**Response body** (`application/json`)

```json
{
  "status": "accepting",
  "enabled": true,
  "connections": 1842,
  "cluster_avg": 1654.33
}
```

| Field | Type | Description |
|-------|------|-------------|
| `status` | string | `"accepting"` or `"rejecting"`. Mirrors the HTTP status code. |
| `enabled` | bool | Whether `balance_enabled` is `on`. When `false`, `status` is always `"accepting"`. |
| `connections` | integer | Active MQTT + MQTT-WS connections on this node. |
| `cluster_avg` | float | Average connection count across all cluster-ready nodes (full precision). |

**Authentication**

Configured via `http_modules_auth` like other HTTP modules. Supports `apikey` (scope `"balance"`) and `noauth`.

## Behavior

- **Disabled by default.** `balance_enabled = off` ships unchanged. Endpoint still responds 200 so existing health checks aren't broken.
- **Fail-open.** If the balance gen_server is unavailable (timeout, crash, not yet started), the endpoint returns 200. A readiness probe never wedges traffic because of a bug in the balance subsystem itself.
- **Skip conditions** (always 200, even when enabled):
  - Empty node (`local_count == 0`): an idle node is never overloaded, regardless of cluster average.
  - Single-node clusters (no peers to redirect to).
  - Fewer than `balance_min_visible_nodes` peers visible this tick (optional tightening knob; default 1 is a no-op).
  - Total cluster connections below `balance_min_connections` (no point balancing in dev/small deployments).
- **Hysteresis to prevent flapping.**
  - Currently accepting: starts rejecting when `local > avg * threshold`.
  - Currently rejecting: resumes accepting when `local < avg * (threshold - hysteresis)`.
- **Hot-reload.** Config changes via `vmq-admin set` take effect on the next check interval; no restart required.
- **Cross-node sampling.** Each tick fans out `vmq_ranch_sup:active_mqtt_connections/0` via `rpc:multicall` with a 2s timeout. Unresponsive nodes are logged at DEBUG, omitted from the average, and counted in `balance_rpc_failures_total`.
- **Jittered ticks.** Each scheduled tick lands within +/- 10 % of the configured `balance_check_interval` to prevent aligned RPC fan-outs across a freshly-restarted cluster.
- **Strict config validation.** Invalid values are rejected at boot rather than silently falling back to defaults.

## New configuration settings

| Setting | Default | Validation | Description |
|---------|---------|------------|-------------|
| `balance_enabled` | `off` | flag (`on`/`off`) | Master switch. When `off`, the endpoint always reports `accepting` and the metric `balance_is_accepting` stays at 1. |
| `balance_threshold` | `"1.2"` | decimal string >= 1.0 | Node stops accepting when `local_count > cluster_avg * threshold`. `1.2` means "20% above average". Must be >= 1.0; values below would reject as soon as a node is at the average. |
| `balance_hysteresis` | `"0.1"` | decimal string >= 0 | Gap below the threshold required to resume accepting. With defaults, a rejecting node resumes when `local < avg * 1.1`. Prevents 200/503 flapping. |
| `balance_min_connections` | `100` | non-negative integer | Skip balancing entirely until total cluster connections reach this number. Avoids balancing in dev/small deployments. |
| `balance_check_interval` | `5000` | integer >= 100 (ms) | How often (in milliseconds) to gather counts and recompute the readiness state. Minimum is enforced to prevent excessive cross-cluster RPC traffic. Actual ticks include +/- 10 % jitter. |
| `balance_min_visible_nodes` | `1` (hidden) | positive integer | Optional tightening: minimum number of cluster-ready nodes this node must see (including itself, via RPC) before it will consider rejecting. Default 1 is a no-op. Raise to 2+ in flaky-network setups to require corroboration before flipping the readiness signal. Below the threshold the node fails open (accepts). Not a quorum requirement. |

All settings are runtime-clamped to safe bounds as a second line of defense against `application:set_env` bypasses.

## New Prometheus metrics

All gauges, on every node, regardless of `balance_enabled`.

| Metric | Type | Description |
|--------|------|-------------|
| `balance_is_accepting` | gauge | `1` if this node is currently accepting new connections, `0` if rejecting due to balance. Always `1` when `balance_enabled = off`. |
| `balance_local_connections` | gauge | Active MQTT + MQTT-WS connections on this node, as seen by the balance subsystem. |
| `balance_cluster_avg` | gauge | Average connection count across cluster-ready nodes. Rounded to integer for the Prometheus surface (the `/api/balance-health` body has full float precision). |
| `balance_is_enabled` | gauge | `1` if cluster auto-balancing is enabled, `0` if disabled. Useful for dashboards/alerts to know whether the rest of these metrics are load-bearing. |
| `balance_rpc_failures_total` | counter | Cumulative count of failed peer count-collection RPCs since node start. Includes both `badrpc` replies (peer reachable but call failed) and nodes the local kernel couldn't reach at all. Steady non-zero rate usually indicates a flaky peer or network. |

## CLI

### `vmq-admin balance show`

Inspect the cluster auto-balance state on the current node. Useful for debugging unexpected `503` responses, network-flakiness investigations, or just sanity-checking that the configured thresholds are what cuttlefish actually applied.

Output is two tables: a settings/state summary and the per-peer connection counts observed on the most recent tick.

```text
+--------------------+---------------+
| Setting            | Value         |
+--------------------+---------------+
| enabled            | true          |
| accepting          | false         |
| local_count        | 30            |
| cluster_avg        | 10.000000     |
| num_visible_nodes  | 3             |
| threshold          | 1.100000      |
| hysteresis         | 0.050000      |
| min_connections    | 0             |
| min_visible_nodes  | 1             |
| check_interval_ms  | 1000          |
| last_tick_ms       | 1778611985003 |
| rpc_failures_total | 0             |
+--------------------+---------------+
+----------------------------+-------------+
| Node                       | Connections |
+----------------------------+-------------+
| VerneMQ@vmq1.cluster.local | 30          |
| VerneMQ@vmq2.cluster.local | 0           |
| VerneMQ@vmq3.cluster.local | 0           |
+----------------------------+-------------+
```

| Field | Description |
|-------|-------------|
| `enabled` | Master switch state. |
| `accepting` | Effective readiness decision the HTTP endpoint will return (`true` always when disabled). |
| `local_count` | Active MQTT + WS connections on this node. |
| `cluster_avg` | Average connection count across cluster-ready nodes (full float precision). |
| `num_visible_nodes` | Distinct cluster-ready peers this node successfully sampled on the last tick, including itself. |
| `threshold` / `hysteresis` | Currently-loaded float values (parsed at boot/runtime from their string config keys). |
| `min_connections` / `min_visible_nodes` / `check_interval_ms` | Currently-loaded integer config values. |
| `last_tick_ms` | Wall-clock timestamp (Erlang `system_time(millisecond)`) of the most recent tick. `undefined` if no tick has run yet. |
| `rpc_failures_total` | Cumulative peer-RPC failures since node start (also exposed as the Prometheus counter `balance_rpc_failures_total`). |

The second table lists what this node observed on the most recent successful tick. Peers that failed to respond are absent (they show up in `rpc_failures_total` instead).

PR #2500, opened 2026-04-07
Add listener address, port, and type to MQTT5 hook

dometec:
## Proposed Changes

I want to apply different authenticazion rules to incoming connections based on the listener the connection is coming from, so I need this info on hook requests.

I ran the tests: 
...
%%% vmq_info_SUITE ==> filtering_works: FAILED
%%% vmq_info_SUITE ==> timetrap_timeout
...
Failed 1 tests. Skipped 10 (10, 0) tests. Passed 501 tests. 

I don't think this is related to my changes.

(Done with support of Cloude Code)

## Types of Changes

## Checklist

- [X] I have read the [CODE_OF_CONDUCT.md](https://github.com/vernemq/vernemq/blob/main/CODE_OF_CONDUCT.md) document
- [X] I have updated changelog (At the bottom of the release version)

### DIFF
--- a/apps/vmq_server/src/vmq_mqtt5_fsm.erl
+++ b/apps/vmq_server/src/vmq_mqtt5_fsm.erl
@@ -68,6 +68,9 @@
         | username()
         | {preauth, string() | undefined},
     conn_opts :: map(),
+    listener_addr :: undefined | tuple() | {local, string()},
+    listener_port :: undefined | non_neg_integer(),
+    listener_type :: undefined | atom(),
     keep_alive :: undefined | non_neg_integer(),
     keep_alive_tref :: undefined | reference(),
     clean_start = false :: flag(),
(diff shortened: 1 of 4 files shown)


ioolkos:
@dometec I welcome this idea in principle. It's not a full implementation of the old plan of allowing different auth plugins per listener. But a good first compromise. There are a couple of subtleties and it needs to be backward compatible... will review the code proposed.
Thanks!

dometec:
Thanks @ioolkos — clarifying the use case:

- Field devices connect via mTLS (HAProxy as LB with SSL offloading) → routed to an MQTT listener
- Mobile app connects via WebSocket → routed to a different listener
- For WebSocket connections, a JWT token is provided in the password field and must be validated

In the hook, I need a way to distinguish these flows so I can decide when JWT validation is required and when it should be skipped.

ioolkos:
@dometec Understood, you need 1 plugin to handle incoming clients differently (based on their listener). A sequential plugin chain is not the same. Specifically, for JWT plugins, if they run in a plugin chain together with DB plugins, one of them will unnecessarily check the password field before passing the request down the chain. Works, but can be confusing.

dometec:
I think I’m missing something about how the plugin chain is supposed to work.
From my understanding, each plugin in the chain processes the same request sequentially.

In my case, the logic is mutually exclusive depending on the listener:

- mTLS devices → no password check needed
- WebSocket clients → password must contain a valid JWT

ioolkos:
@dometec For client certs, it's a bit different. If you want to disable username/password checks for incoming connections (because you take the client cert as proof of identity already), then you have to override the allow_anonymous setting for that listener: `listener.ssl.my_listener.allow_anonymous_override = on`. By that, you switch off authentication for the listener. (Note that the global setting `allow_anonymous = off` needs to stay off, of course).

The WS clients on the other listener will then just be handled by whatever other plugin you have started.

But yes, in general, the request goes sequentially through all plugins until one matches. But it does this in a given order, and if only the last plugin matches you have more delay.

dometec:
Thanks @ioolkos! I didn't realize the existence of allow_anonymous_override on listener... shame on me :)

ioolkos:
No shame, documentation should be better... :sweat_smile:

Issue #2412, opened 2025-06-05
[Bug]: Race condition: message lost when published immediately after SUBACK (ETS update lags)


AlexeyKiselev824:
### Environment

- VerneMQ Version: 2.1.0
- OS: Debian GNU/Linux 12 (bookworm)
- Erlang/OTP version (if building from source):
- Cluster size/standalone: single node and 3-node cluster

## Summary
On VerneMQ 2.1.0 a message published immediately after the client receives SUBACK is sometimes never delivered to the subscriber. This race condition occurs naturally during the production process, but it is difficult to reproduce it without artificial delay.

#### Related discussion: #1760
That thread seems to describe the same underlying problem, but it has been inactive for quite a while, so I’m opening this fresh issue with a minimal, reproducible test case

## Steps to Reproduce
1. VerneMQ (one node is enough)

2. Add artificial delay in `apps/vmq_server/src/vmq_reg_trie.erl` in `handle_event/2` 
   **BEFORE** the ETS table updates:
   ```erlang
   handle_event(Handler, Event) ->
       timer:sleep(1000), % Add this line to simulate async delay
       case Handler(Event) of
       %%  ...
   ```
   This simulates the natural async delay between DB write and ETS update.

3. Build and start VerneMQ with the modified code, ensure MQTT port 1883 is available

4. Run the demo script (in the attachment)

[vernemq-bug-demo.tar.gz](https://github.com/user-attachments/files/20613965/vernemq-bug-demo.tar.gz)

### Docker
```bash
docker-compose up --build
```

### What the script does:
1. **SUBSCRIBE** → wait for SUBACK
2. **PUBLISH** → immediately after SUBACK (QoS 1)  
3. Wait for message delivery

### Actual Result
The message is dropped. Inside `vmq_reg:publish_fold_wrapper/4` no subscriber is found because the ETS table has not yet been updated, even though SUBACK has already been sent.

## Root Cause (code level)

**Subscription flow:**
1. SUBSCRIBE → `vmq_subscriber_db:store` (sync DB write)
2. SUBACK sent immediately after DB write
3. Plumtree generates event → `vmq_reg_trie:handle_event` → ETS update (async)

**Publication flow:** searches ETS tables

## Possible Fixes
From my perspective, while I'm not deeply familiar with Erlang/VerneMQ internals, some potential approaches might include:

1. **Retry mechanism**: Add retry logic in `publish_fold_wrapper` when no subscribers are found
[...]

PR #2504, opened 2026-05-06
Poll the local subs table before confirming subscriptions

ioolkos:
A minimal fix for https://github.com/vernemq/vernemq/issues/2412

This waits for RAM tables to have a subscription stored before the session sends out a SUBACK. Since it uses a polling approach, we'll have to test behaviour under considerable load. The added logs & metrics will allow conclusions.

Note that this only takes care of the local subscription table. Waiting/coordinating in a cluster will be a next step.


ioolkos:
Close in favour of #2514

Issue #1755, opened 2021-03-17
Send listener name in auth_on_register hook for lua script


mtaobiz:
We need listener name in lua auth script to make authentication per listener.

### Environment

- VerneMQ Version: 1.11.0
- OS: Debian 10
- Cluster size/standalone: Standalone

hennejg:
I am currently researching a related topic. I therefore think that this issue is closely related to #1526.

ioolkos:
@hennejg thanks, where's the relation?

hennejg:
I suppose that @mtaobiz trying to achieve something similar to what I am trying to do: set up multiple listeners with different authentication strategies, like one with user/password a second one with client-certificate based authentication. Currently, to the best of my understanding, you can't tell in the listener whether the username is the CN from a certificate or an actual username argument from a username/password authentication attempt. If the hook had more context information about the authentication attempt, ideally access to the full certificate details, that would be rather easy.

The OP's requirement might be slightly different like trying to support username/password authentication against different backends. Then the relation is a bit more distant, but the requirement could still be subsumed into a larger "need more context info in auth_on_register" feature request. 
However, some discussion in #1526 centered around adding a hook specifically for certificate verification. If that approach would be chosen, it would do nothing for _this_ feature request.

tradingpanel:
set up multiple listeners with different authentication strategies

mtaobiz:
set up multiple listeners with different authentication strategies

ioolkos:
Jep, that'd be a nice & useful feature.

siegerhansma:
We're running into the same problem. For some context, our use case:

Some clients connect with username and password and are routed to 1883 (with TLS termination done on the proxy side). Other client connect to 8883 and have use a certificate to connect, with the CN used as username. In the `auth_on_register` hook, we can do a check if it is the correct user. 

But because we can't differentiate between the 1883 and 8883 listener, we don't know if a user connects with a missing password because they didn't enter the password or because it is using the certificate.

siegerhansma:
After posting the comment, I just saw https://github.com/vernemq/vernemq/pull/2511 being worked on, which would help with this.

ioolkos:
@siegerhansma Thanks! #2511 doesn't add the feature for Lua scripts (vmq_diversity) currently. Note that I've also just added a related feature for per-listener plugin chains: https://github.com/vernemq/vernemq/pull/2520, but we'll need to verify and maybe adapt it for Lua scripts as well.

---
[...]

PR #2511, opened 2026-06-22
Add listener address, port, and type to MQTT5 hook

dometec:
Reopened from #2500 (wrong rebase...)

## Proposed Changes

I want to apply different authenticazion rules to incoming connections based on the listener the connection is coming from, so I need this info on hook requests.

I ran the tests:
...
%%% vmq_info_SUITE ==> filtering_works: FAILED
%%% vmq_info_SUITE ==> timetrap_timeout
...
Failed 1 tests. Skipped 10 (10, 0) tests. Passed 501 tests.

I don't think this is related to my changes.

(Done with support of Cloude Code)

## Types of Changes

## Checklist

- [X] I have read the [CODE_OF_CONDUCT.md](https://github.com/vernemq/vernemq/blob/main/CODE_OF_CONDUCT.md) document
- [X] I have updated changelog (At the bottom of the release version)


ioolkos:
@dometec This PR is just an edit of the changelog. Feel free to comment on https://github.com/vernemq/vernemq/pull/2520 to see whether this covers your needs or not. (Your original intention might have been a bit different, though)

dometec:
I've added the missing commit.

Yes, #2520 is the right solution for selecting authentication and authorization plugin chains on a per-listener basis. However, I would still like to include this PR as well, since it provides additional listener context to hook calls, which may also be useful for logging, auditing, and plugin-specific logic.

ioolkos:
@dometec You meant to have this for both v5 and v3, at least according to the changelog line?

I also think there's something to think about regarding arity 7 and 6 versions of the hooks. We need to be able to run mixed chains of plugins. Need to check if that's still guaranteed.

dometec:
Yes, the intention is to support both MQTT v3 and MQTT v5. I'm currently implementing the same changes for MQTT v3 as well.

Regarding the hook arities, thanks for pointing that out. I'll review that part carefully to make sure mixed chains of arity 6 and arity 7 hooks continue to work correctly.

I also really appreciate the feedback. I'm new to Erlang, so comments like these are very helpful and help me better understand the existing design.

dometec:
done!

ioolkos:
@dometec Cool! We can make this even better if we actually make it configurable. The main reason is that existing old plugins might not even implement auth_on_register/6, so the unconditional injection of listener opts will break them.

But I realised we have a `forward_connection_opts` flag for TLS/WSS listeners, to make client cert info forward to plugins. If we extend the same `forward_connection_opts` flag to TCP/WS listeners, we have a global per-listener flag for your feature. The only "surprise" is then for existing SSL listeners/plugins that your listener info is added as well. But I think that's acceptable in the name of having a simple setting.

So, in short:
- When `forward_connection_opts = false`: call legacy auth hook arities only.
- When `forward_connection_opts = true`: call extended auth hook arities with a map of available connection metadata.
- For TLS/WSS: include client_cert plus listener metadata.
- For TCP/WS: include listener metadata only.

Check for the setting can still happen in `vmq_ranch_config:default_session_opts/1`. With something along the lines of  

```erlang
MaybeForwardConnectionOpts =
    case proplists:get_value(forward_connection_opts, Opts, false) of
        true ->
            [
                {listener_addr, ListenerAddr},
                {listener_port, ListenerPort},
                {listener_type, ListenerType}
            | PreviousDefaultOpts];
        false ->
            PreviousDefaultOpts
```

Let me know if you see any cons for that idea, of course.

dometec:
Thanks for the suggestion — I’ve updated the PR accordingly.

forward_connection_opts now controls which authentication hook arity is called:
When disabled, only the legacy hook arities are used.
When enabled, the extended arities receive the available connection metadata.

TCP/WS listeners receive listener metadata.
TLS/WSS listeners also receive the client certificate.

I also added coverage for clear TCP connections with Proxy Protocol enabled and fixed an issue where the code could attempt to retrieve a TLS certificate from a plain TCP socket.

I’m moving this PR back to draft because I’d like to run some additional manual tests over the next few days. In the meantime, any comments or observations are very welcome.

ioolkos:
@dometec Thanks, looks like the feature is complete to me. Let me know when your manual testing passes your requirements & I can add this to the release I'm preparing.

dometec:
rebased, test ok, thanks!

apps/vmq_webhooks/src/vmq_webhooks_plugin.erl
```diff
@@ -354,7 +374,20 @@ auth_on_register_m5(Peer, SubscriberId, UserName, Password, CleanStart, Props, O
...
+            {local, _} -> <<"local">>;
+            LAddr when is_tuple(LAddr) -> list_to_binary(inet:ntoa(LAddr))
+        end,
+    ListenerPort = nullify(maps:get(listener_port, Opts, undefined)),
+    ListenerType =
+        case maps:get(listener_type, Opts, undefined) of
+            undefined -> null;
+            LT -> atom_to_binary(LT, utf8)
+        end,
+    FilteredOpts = maps:without([listener_addr, listener_port, listener_type], Opts),
```
ioolkos: This section needs rework. BaseArgs should not be the full list. You can also not ++ a list and a map: BaseArgs ++ [FilteredOpts]. Same applies to the `auth_on_register` clause of the function.
This also shows we don't have test coverage for forward_connection_opts in the webhooks plugin, since the webhooks SUITE passes.

ioolkos (changes requested):

dometec:
Ok, I think now is ok. Add also some other unit tests.

ioolkos:
@dometec I added the feature for vmq_diversity as well, since it was requested (https://github.com/vernemq/vernemq/issues/1755#issuecomment-5178239484, cc @siegerhansma )
I also refactored the opts handling a bit, so that we could use the hook Opts for future features independent of `forward_connection_opts`. Listener metadata is only added when `forward_connection_opts` is enabled, of course.

Let me know, if this passes your local tests (and requirements)

dometec:
LGTM, thanks!

Issue #1369, opened 2019-11-19
got unknown sync event in drain state


chenxinlong:
### Environment

- VerneMQ Version:1.7.1
- OS: CentOS 7 
- Cluster size/standalone: cluster with 5 nodes (docker) 

At the time between 2019-11-18 18:20(UTC) and 2019-11-18 18:40(UTC),  there's a lot of error outputs in the error.log of one of my nodes: 
```
2019-11-18 18:21:51.256 [error] <0.10418.0>@vmq_queue:wait_for_offline:280 got unknown event in wait_for_offline state {change_state,active,<0.24261.30>}
2019-11-18 18:22:31.125 [error] <0.11068.20>@vmq_queue:drain:428 got unknown sync event in drain state {add_session,<0.32090.30>,#{allow_multiple_sessions => false,cleanup_on_disconnect => false,is_plugin => false,max_drain_time => 500,max_msgs_per_drain_step => 100,max_offline_messages => 1000,max_online_messages => 1000,queue_deliver_mode => fanout,queue_type => fifo}}
2019-11-18 18:22:31.286 [error] <0.11068.20>@vmq_queue:drain:428 got unknown sync event in drain state {add_session,<0.32090.30>,#{allow_multiple_sessions => false,cleanup_on_disconnect => false,is_plugin => false,max_drain_time => 500,max_msgs_per_drain_step => 100,max_offline_messages => 1000,max_online_messages => 1000,queue_deliver_mode => fanout,queue_type => fifo}}
2019-11-18 18:22:31.506 [error] <0.11068.20>@vmq_queue:drain:428 got unknown sync event in drain state {add_session,<0.32090.30>,#{allow_multiple_sessions => false,cleanup_on_disconnect => false,is_plugin => false,max_drain_time => 500,max_msgs_per_drain_step => 100,max_offline_messages => 1000,max_online_messages => 1000,queue_deliver_mode => fanout,queue_type => fifo}}
2019-11-18 18:22:31.904 [error] <0.11068.20>@vmq_queue:drain:428 got unknown sync event in drain state {add_session,<0.32090.30>,#{allow_multiple_sessions => false,cleanup_on_disconnect => false,is_plugin => false,max_drain_time => 500,max_msgs_per_drain_step => 100,max_offline_messages => 1000,max_online_messages => 1000,queue_deliver_mode => fanout,queue_type => fifo}}
```

At the same time, the load average of this node reached an unusually high value, like the pic below : 

[...]

Issue #571, opened 2018-01-08
error: unknown event in wait_for_offline state


michaelfung:
### Environment

- VerneMQ Version:  commit a97b351
- OS: both ubuntu 17.10
- Erlang/OTP version (if building from source):  20.0.4

The client was disconnected by the server. The client side error is:
```
Error: Connection reset by peer 
```

The broker had the following in the log:
```
2018-01-07 18:17:43.370 [error] <0.455.0>@vmq_queue:wait_for_offline:254 got unknown event in wait_for_offline state {change_state,notify,<0.451.0>}
```

Is this a broker side problem?

larshesel:
Hi @michaelfung 

The log message above means that there's an unexpected event happening while the queue is shutting down. It may or may not be an actual problem.

I suppose you experience an unexpected disconnect from the server? Can you tell us more about what your configuration is and what is happening when this occurs?

michaelfung:
Yes, the disconnect was unexpected. It only happened a few times so far.

I am writing a perl script to stress test my system. It uses AnyEvent::MQTT to make thousands of connections to the broker and do periodic publishing. The configuration file is almost default with these web hooks configurations:
```
vmq_webhooks.hub_auth.hook = auth_on_register
vmq_webhooks.hub_auth.endpoint = http://localhost:3000/hooks/auth_on_register

vmq_webhooks.hub_register.hook = on_register
vmq_webhooks.hub_register.endpoint = http://localhost:3000/hooks/on_register

vmq_webhooks.hub_sub_auth.hook = auth_on_subscribe
vmq_webhooks.hub_sub_auth.endpoint = http://localhost:3000/hooks/auth_on_subscribe

vmq_webhooks.hub_pub_auth.hook = auth_on_publish
vmq_webhooks.hub_pub_auth.endpoint = http://localhost:3000/hooks/auth_on_publish
```

Nothing unusual observed at the webhooks log files. 

It also can be the AnyEvent::MQTT library at fault. It would be helpful if vmq show more details in the error log.

PR #2524, opened 2026-07-31
Fix session-takeover wedge: handle change_state in wait_for_offline (#571, #1369)

jessequinn:
## Problem

`vmq_queue` can wedge permanently in `wait_for_offline` during a session takeover, after which every subsequent CONNECT for that client_id is rejected with `multiple sessions not allowed` until the broker is restarted. This matches the long-standing reports in #571 and #1369.

### Root cause

On takeover (`allow_multiple_sessions=false`), `online/3` disconnects the old session(s), stores the new CONNECT as `waiting_call`, and transitions `online -> wait_for_offline`; the takeover completes only when the old session goes `'DOWN'`.

Session processes report their own state asynchronously via `vmq_queue:active/1` and `notify/1`, casting `{change_state, active|notify, Pid}`. `online/2` handles this event, but **`wait_for_offline/2` had no clause for it**, so the state's catch-all logged `got unknown event in wait_for_offline state ...` and dropped it.

If a session that was told to disconnect **re-activates instead of dying** (it flaps — e.g. after a "Connection reset by peer"), it emits `change_state` rather than `'DOWN'`. Pre-fix, that event was swallowed, the session was never re-disconnected, never went `'DOWN'`, the `waiting_call` was never replied, and the queue stayed wedged in `wait_for_offline` indefinitely.

## Fix

Add a `wait_for_offline/2` clause for `change_state`:

- For a **tracked** session: update bookkeeping via `change_session_state/3` (as `online/2` does) and re-issue the disconnect matched to the current `waiting_call` (new `disconnect_reason/1` helper: `add_session -> ?SESSION_TAKEN_OVER`, `migrate -> ?DISCONNECT_MIGRATION`, `{cleanup, R} -> R`), driving the session back toward `'DOWN'` so the pending call completes.
- For an **untracked** pid (the pending replacement session held in `waiting_call`, not yet in the `sessions` map): ignore it, because `change_session_state/3` does `maps:get/2` and would otherwise crash.

Tradeoff (noted in code): a session that keeps ignoring disconnects could re-trigger the clause, but each pass sends another disconnect and drives toward `'DOWN'` — strictly better than the previous swallow-and-wedge.

## Test

Adds `queue_wait_for_offline_change_state_test` to `vmq_queue_SUITE`, which drives the exact flap-during-takeover sequence with a `flapping_session/3` helper (re-activates once on the takeover disconnect, then dies) and asserts the takeover still completes. Pre-fix it times out (wedged); post-fix it completes.

## Notes for reviewers

- The `change_session_state/3` call in the new clause mirrors `online/2`; if you'd prefer the minimal variant (re-disconnect only, no bookkeeping update) I'm happy to adjust.
- I authored this against a 2.1.1 deployment and verified the edited regions are byte-identical on `main`; it rebases cleanly here. I don't have an Erlang/rebar3 toolchain in my authoring environment, so I've **not run `rebar3 compile`/`ct`/`dialyzer` locally** — relying on CI here. Happy to iterate on anything.

### DIFF
--- a/apps/vmq_server/src/vmq_queue.erl
+++ b/apps/vmq_server/src/vmq_queue.erl
@@ -311,6 +311,22 @@ online(Event, _From, State) ->
     ?LOG_ERROR("got unknown sync event in online state ~p", [Event]),
     {reply, {error, online}, State}.
 
+%% A session we are draining can report a state change ({change_state, ...} via
+%% active/1 or notify/1) instead of going 'DOWN'; handle it here so the catch-all
+%% doesn't swallow it and wedge the takeover in wait_for_offline (#571, #1369).
+%% Re-disconnect a tracked session; ignore an untracked pid (the pending
+%% replacement in waiting_call) so we don't disconnect the newcomer.
+wait_for_offline({change_state, _NewSessionState, SessionPid}, State) ->
+    case maps:is_key(SessionPid, State#state.sessions) of
+        true ->
+            vmq_mqtt_fsm_util:send(
+                SessionPid,
+                {disconnect, disconnect_reason(State#state.waiting_call)}
+            ),
+            {next_state, wait_for_offline, State};
+        false ->
+            {next_state, wait_for_offline, State}
+    end;
 wait_for_offline({enqueue, Msg}, State) ->
     _ = vmq_metrics:incr_queue_in(),
     {next_state, wait_for_offline, insert(Msg, State)};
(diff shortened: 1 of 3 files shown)


apps/vmq_server/src/vmq_queue.erl
```diff
@@ -311,6 +311,41 @@ online(Event, _From, State) ->
...
+%% in waiting_call, which is NOT yet in the sessions map -- is ignored, because
+%% change_session_state/3 does maps:get/2 and would crash on an unknown pid.
+%%
+%% Tradeoff: a pathological session that keeps ignoring disconnects could keep
+%% re-triggering this clause, but every pass sends another disconnect and drives
+%% toward 'DOWN'; this is strictly better than the previous swallow-and-wedge.
+wait_for_offline({change_state, NewSessionState, SessionPid}, State) ->
+    case maps:is_key(SessionPid, State#state.sessions) of
+        true ->
+            NewState = change_session_state(NewSessionState, SessionPid, State),
```
ioolkos: I'd not run change_session_state at all (potential to send *more* messages to the disconnected session)

ioolkos (changes requested):
Thanks! Can you address the comment on line 340, and delete/reduce the commentary? 
(I think some of the wording might be a bit misleading... like flapping)

ioolkos:
@jessequinn Overall, I think this is a real bugfix! I've added some comments in my review. Forgot to mention: could you also add a line to the changelog?

apps/vmq_server/src/vmq_queue.erl
jessequinn: Good catch, thanks — done in c9573a2a. Dropped the `change_session_state/3` call so the clause now just re-issues the takeover disconnect. The `maps:is_key/2` guard still ensures we only ever disconnect a session we're already draining, never the pending replacement session in `waiting_call`. Comment updated to reflect the reasoning.

jessequinn:
Thanks for the review! Addressed in c9573a2a + 73638637:

- Dropped `change_session_state/3` in the `wait_for_offline` clause — it now just re-issues the takeover disconnect (line 340).
- Trimmed the comment way down (~24 → 5 lines) and removed the misleading flap / "connection reset" wording.
- Added a CHANGELOG bugfix line.

Happy to trim further or adjust anything else.

jessequinn:
Updated in e7b7be7c:
- Trimmed the `vmq_queue_SUITE.erl` commentary as well, and renamed the `flapping_session` test helper to `reactivating_session` to drop the misleading wording.
- Rebased onto latest `main` (resolved the `changelog` conflict — kept your new entries plus the bugfix line) and squashed to a single commit.

Should be conflict-free now.

PR #2525, opened 2026-08-02
Fix/queue migration message loss

ioolkos:


ioolkos:
Note to self: looks like this is unfinished, as we'd have to resolve mixed-cluster compatibility (for rolling upgrades).

